PR #4, opened 2022-02-05
add support for empty nd-array syntax

simeonschaub:
What's still missing here is the correct compat check, since this is
only supported in Julia 1.8. I wasn't sure how to check for this in
`check_ncat_compat`.

### DIFF
--- a/src/parser.jl
+++ b/src/parser.jl
@@ -2725,6 +2725,12 @@ function parse_cat(ps::ParseState, closer, end_is_symbol)
     if k == closer
         # []  ==>  (vect)
         return parse_vect(ps, closer)
+    elseif k == K";"
+        # [;;]      ==>  (ncat 2)
+        # [;; \n ]  ==>  (ncat 2)
+        n_semis, _ = parse_array_separator(ps)
+        bump_closing_token(ps, closer)
+        return (K"ncat", set_numeric_flags(n_semis))
     end
     mark = position(ps)
     parse_eq_star(ps)
--- a/src/parser.jl
+++ b/src/parser.jl
@@ -2673,13 +2673,13 @@ end
 #   power of 0 for whitespace and negative numbers for other separators.
 #
 # FIXME: Error messages for mixed spaces and ;; delimiters
-function parse_array_separator(ps)
-    t = peek_token(ps)
+function parse_array_separator(ps; skip_newlines=false)
+    t = peek_token(ps; skip_newlines=skip_newlines)
     k = kind(t)
     if k == K";"
         n_semis = 1
         while true
-            bump(ps, TRIVIA_FLAG)
+            bump(ps, TRIVIA_FLAG; skip_newlines=skip_newlines)
             t = peek_token(ps)
             if kind(t) != K";" || t.had_whitespace
                 break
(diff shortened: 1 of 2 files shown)


c42f (commented):
Awesome, this is the first PR to this repo!

For now, this will need the tests like `[;;]      ==>  (ncat 2)` added in test/parser.jl  (actually this should be `(ncat-2)` because the `2` is stored in the flags, not as a child node in the tree.  Search for similar tests to figure out where best to add them.

For interactively developing the parser and tests, I'd suggest using the `itest_parse` function which will check several things and do some pretty printing:

```julia
julia> include("test/test_utils.jl")
julia> itest_parse(JuliaSyntax.parse_atom, "[;;]")

# Code:
[;;]

# Green tree:
     1:4      │[error]                   ✘
     1:4      │  [ncat]
     1:1      │    [                        "["
     2:2      │    ;                        ";"
     3:3      │    ;                        ";"
     4:4      │    ]                        "]"
Error: unexpected closing token:
[;;]
Error: multidimensional array syntax not supported in Julia version 1.6 < 1.7:
[;;]

# SyntaxNode:
(error (ncat-2))

# Julia Expr:
:($(Expr(:error, :([;;]))))

# AST dump
Expr
  head: Symbol error
  args: Array{Any}((1,))
    1: Expr
      head: Symbol ncat
      args: Array{Any}((1,))
        1: Int64 2

# flisp Julia Expr:
:($(Expr(:error, "unexpected \";\"")))

```

Also, optional... but perhaps consider moving this logic to the top of `parse_array`, because that's the function which handles the rest of `ncat` syntax?

> I wasn't sure how to check for this in `check_ncat_compat`.

This should already work as-is, because you emit `K"ncat"` and `check_ncat_compat` just checks for that.

c42f:
> This should already work as-is, because you emit K"ncat" and check_ncat_compat just checks for that.

Oh, I see the problem... it's not the same version check as other `ncat` stuff which is in 1.7.

We could consider adding the version check immediately where you detect the `K";"`? The only issue with that is it will cause two related and somewhat redundant errors on Julia <= 1.6, one for ncat syntax and one for this particular empty array syntax. But perhaps we can live with that.

c42f:
> just check if the ncat expression is otherwise empty in check_ncat_compat, but I wasn't sure how to do that

We'd have to add more sophisticated ways of looking back on the output to make that easy. That's possible, but I do like to avoid it because I think it leads to confusing code. I'd even like to remove `peek_behind()` completely but I'm not sure it's possible...

src/parser.jl
```diff
@@ -2722,11 +2722,18 @@ function parse_cat(ps::ParseState, closer, end_is_symbol)
                     whitespace_newline=false,
                     for_generator=true)
     k = peek(ps, skip_newlines=true)
+    mark = position(ps)
     if k == closer
         # []  ==>  (vect)
         return parse_vect(ps, closer)
+    elseif k == K";"
+        # [;;]      ==>  (ncat 2)
+        # [;; \n ]  ==>  (ncat 2)
```
c42f, line 2731: I'll fix these separately to keep them in sync with the actual tests (yeah, I need to improve the tooling so we don't need to repeat ourselves :sweat_smile: )

c42f:
This looks excellent, thanks :-D

src/parser.jl
simeonschaub: Oops! Yeah, I think I was just thinking of the s-expression syntax instead of JuliaSyntax' one.

PR #6, opened 2022-02-14
Compat for odd Exprs in long form anonymous functions

c42f:
The following construct is parsed oddly by the reference parser with the
tuple omitted from the argument list. Add a compatibility hack for that
during Expr conversion.

```julia
    function (xs...)
        body
    end
```

With this we now have compatible parsing for all of the stdlib, so turn
on this test by default.


c42f:
> With this we now have compatible parsing for all of the stdlib, so turn on this test by default.

Not quite true for 1.8 yet, so this is disabled conditionally for now.

PR #15, opened 2022-02-23
get rid of some unused code in tokenize

KristofferC:


c42f (approved):
I think let's merge this as it's (presumably) ancient code.

Though at some point I'd like to replicate the logic from `Base.is_id_start_char()` and not depend on the Base C code for this so we may need to bring a modernized version of this back.  Why? Well I want JuliaSyntax to parse syntax from various Julia versions independently of the Julia version it's hosted on (this makes it more usable in language server).

pfitzseb:
Re `is_id_start_char`: We *could* use part of this [beautiful regexp](https://gist.github.com/pfitzseb/1e5cb6c52b196bb556983959c1aef02b) for that. Probably better to just translate the C source into Julia though.

c42f:
Oh wow ... that's quite a regex!

PR #16, opened 2022-02-23
Fold `RawToken` into `Token` and remove the old `Token`.

KristofferC:
I think it is best to go "all in" on that Token only stores the byte ranges. No one will use the Token that stores the string explicitly because it is too slow. This allows quite a lot of code removal.


Tokenize/src/token.jl
```diff
@@ -82,98 +62,31 @@ struct RawToken <: AbstractToken
...
 
-function Base.show(io::IO, t::Token)
-    start_r, start_c = startpos(t)
-    end_r, end_c = endpos(t)
-    str = kind(t) == ENDMARKER ? "" : escape_string(untokenize(t))
-    print(io, rpad(string(start_r, ",", start_c, "-", end_r, ",", end_c), 17, " "))
-    print(io, rpad(kind(t), 15, " "))
-    print(io, "\"", str, "\"")
+function untokenize(t::Token, str::String)
+    String(codeunits(str)[1 .+ (t.startbyte:t.endbyte)])
```
c42f, line 86: This is a nice way to express this operation with `codeunits`.  I'm usually messing with `previndex` for this.

Tokenize/test/lexer.jl
```diff
@@ -296,205 +296,220 @@ end
 @testset "show" begin
     io = IOBuffer()
     show(io, collect(tokenize("\"abc\nd\"ef"))[2])
-    @test String(take!(io)) == "1,2-2,1          STRING         \"abc\\nd\""
+    @test String(take!(io)) == "1-5        STRING         "
 end
 
-~(tok::T.AbstractToken, t::Tuple) = tok.kind == t[1] && untokenize(tok) == t[2]
+~(tok::T.Token, t::Tuple) = tok.kind == t[1] && untokenize(tok, t[3]) == t[2]
```
c42f, line 302: Oho. I see what you did there.  Very tricky.

Maybe neater to just define a utility function to take the place of `collect(tokenize(ts))` instead which yields the tuples I'm testing against below. Then use `==` rather than having to pass `str` everywhere.

(Nevertheless, let's merge this as-is; that could be cleaned up separately.)

c42f (approved):
What an unexpected-but-welcome gift of code.  Thanks @KristofferC !

PR #17, opened 2022-02-23
get rid of the `TokenError` field in `Token`

KristofferC:
It is very rare for tokens to be errors but we still need to pay the cost of the `TokenError` field every time. This PR just moves the `TokenError` into a `Kind` and has a function `iserror` to check if the kind is of an error type.

Made on top of https://github.com/c42f/JuliaSyntax.jl/pull/16

### DIFF
--- a/src/parse_stream.jl
+++ b/src/parse_stream.jl
@@ -55,7 +55,7 @@ function Base.summary(head::SyntaxHead)
 end
 
 function untokenize(head::SyntaxHead; unique=true, include_flag_suff=true)
-    str = untokenize(kind(head); unique=unique)
+    str = is_error(kind(head)) ? "error" : untokenize(kind(head); unique=unique)
     if is_dotted(head)
         str = "."*str
     end
--- a/README.md
+++ b/README.md
@@ -405,7 +405,7 @@ name of compatibility, perhaps with a warning.)
   broken-looking AST like `(macrocall (. A (quote (. B @x))))`.  It should
   probably be rejected.
 * Operator prefix call syntax doesn't work in the cases like `+(a;b,c)` where
-  keyword parameters are separated by commas. A tuple is produced instead. 
+  keyword parameters are separated by commas. A tuple is produced instead.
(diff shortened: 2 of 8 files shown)


KristofferC:
Rebased

src/parse_stream.jl
```diff
@@ -55,7 +55,7 @@ function Base.summary(head::SyntaxHead)
 end
 
 function untokenize(head::SyntaxHead; unique=true, include_flag_suff=true)
-    str = untokenize(kind(head); unique=unique)
+    str = is_error(kind(head)) ? "error" : untokenize(kind(head); unique=unique)
```
c42f, line 58: It seems a potentially dubious to special case this in here rather than in `untokenize(::Kind)`. I assume something fails somehow without this?

c42f (commented):
Looks good. I think `is_error(head::SyntaxHead)` may need fixing after this, and I feel slightly worried by the special case in `untokenize`.

Somewhat unrelated — you'll notice that the way I've bridged the parser with the lexer is hardly ideal in terms of repetition and consistency, especially for kinds. I hope to clean this up at some point by changing the way kinds are declared, probably moving away from `@enum` and toward the string macro approach. When parsing, it's extremely helpful to have the kind names be self representing (ie, the kind of `for` is `K"for"`).

src/parse_stream.jl
KristofferC: Right now, the parser expects any error to `untokenize` to `"error"`. But now we have many error kinds. I could just add all of them to https://github.com/c42f/JuliaSyntax.jl/blob/77b4044218a7c7c5e34f8b1459a88fe6d405b64c/src/token_kinds.jl#L6 and have them sting to `"error"`.

c42f: Yes I think they should be added to `_str_to_kind` — this allows them to be accessed with the `K` macro.

c42f:
I think let's merge this. I'll add some cleanup to to `is_error` and `untokenize()` in a follow-up.

PR #18, opened 2022-03-04
Optimization and cleanup in parse stream

c42f:
This is a series of cleanup/optimization commits which

* Reduce the size of data structures in `ParseStream`
* Fix various JET warnings, particularly removing boxed closure captures and fixes to `EMPTY_TOKEN` in Tokenize
* Pool and reuse some data structures during parsing
* Better kind()/flags() API + compactify SyntaxToken flags
* Remove need for storing newline flag on SyntaxToken


c42f:
Overall this seems to boost the parser's speed by around 40% :-)

PR #22, opened 2022-03-10
Prototype for conversion to CSTParser.EXPR

c42f:
This demonstrates the basic approach working, but various things don't
work here yet due to two things:
* Mismatch between the way the two packages tokenize the input, eg, for
  the delimiters in strings.
* Missing "trivia nodes" in JuliaSyntax (eg, a brackets node). These
  should probably be added.


c42f:
@pfitzseb here's an initial prototype of the `CSTParser.EXPR` conversion.

Currently it's still in the prototypes directory which I think is fine while experimenting.

However to make this usable but also avoid a `CSTParser` dependency we could either
* Put this in a separate compatibility package.
* Duplicate the `CSTParser.EXPR` data structure, in a `JuliaSyntax.CSTParser` compat module. But we'd also need to duplicate some CSTParser predicates and potentially the iteration code which could add up to a lot of code.

davidanthoff:
Another option would be that CSTParser.jl takes a dependency on JuliaSyntax.jl, and just starts to use the parser her under the hood? We might even be able to keep the complete CSTParser API unchanged and non-breaking?

c42f:
Sure! If everyone maintaining CSTParser is happy to do that, I think that's an excellent solution!

sjkelly:
I suppose this would be even cleaner if CSTParser.jl incorporated this code as a package extension?

davidanthoff:
I think it would still be cleaner to go with my idea from above. We have had serious difficulty maintaining even just one parser in CSTParser due to a lack of maintainer time, so if we were to go with a package extension story, we would effectively have to maintain two parallel things at the same time, which would presumably require even more time. Plus, I don't really see any benefit of that?

Issue #31, opened 2022-06-23
Some tokenize issues


c42f:
Running the tests from the system image, I found some broken tests due to operators which should be errors, but the JuliaSyntax lexer tokenizes them as operators:

```
julia> broken_ops = [
        "a .-> b",
        "a .>: b",
        "a .<: b",
        "a ||₁ b",
        "a ||̄ b",
        "a .||₁ b",
        "a &&₁ b",
        "a &&̄ b",
        "a .&&₁ b",
       ]
9-element Vector{String}:
 "a .-> b"
 "a .>: b"
 "a .<: b"
 "a ||₁ b"
 "a ||̄ b"
 "a .||₁ b"
 "a &&₁ b"
 "a &&̄ b"
 "a .&&₁ b"

julia> [[JuliaSyntax.Tokenize.untokenize(t, s) for t in JuliaSyntax.Tokenize.tokenize(s)] for s in broken_ops]
9-element Vector{Vector{String}}:
 ["a", " ", ".->", " ", "b", ""]
 ["a", " ", ".>:", " ", "b", ""]
 ["a", " ", ".<:", " ", "b", ""]
 ["a", " ", "||₁", " ", "b", ""]
 ["a", " ", "||̄", " ", "b", ""]
 ["a", " ", ".||₁", " ", "b", ""]
 ["a", " ", "&&₁", " ", "b", ""]
 ["a", " ", "&&̄", " ", "b", ""]
 ["a", " ", ".&&₁", " ", "b", ""]
```

pfitzseb:
Dotted sub/supertype ops parse fine though:
```
julia> Meta.parse("a .<: b")
:(a .<: b)

julia> Meta.parse("a .>: b")
:(a .>: b)

julia> dump(Meta.parse("a .>: b"))
Expr
  head: Symbol call
  args: Array{Any}((3,))
    1: Symbol .>:
    2: Symbol a
    3: Symbol b
```

c42f:
Good point! I extracted those wrongly. The problem with those operators is for the unary versions `.<:` and `.>:`

```
julia> JuliaSyntax.parseall(Expr, ".<: b", rule=:statement)
:(<:b)

julia> Meta.parse(".<: b")
ERROR: Base.Meta.ParseError("\".<:\" is not a unary operator")
Stacktrace:
...

julia> JuliaSyntax.parseall(Expr, ".>: b", rule=:statement)
:(>:b)

[...]

PR #32, opened 2022-06-23
Fix some dotop tokenization issues

pfitzseb:
Fixes https://github.com/JuliaLang/JuliaSyntax.jl/issues/31.


c42f (approved):
Excellent, thanks very much :-)

I streamlined the tests a bit. Let's merge, provided CI likes it.

I also thought about adding the token kind to the tests, as is done in some other places.  But that made the tests much more verbose and hard to understand, so I left it out out in the end.

PR #37, opened 2022-08-05
Separate UUID for the JuliaSyntax in sysimage

c42f:
This allows us to use JuliaSyntax baked into the sysimage to develop
a separate JuliaSyntax with Revise :-D

### DIFF
--- a/sysimage/JuliaSyntaxCore/Project.toml
+++ b/sysimage/JuliaSyntaxCore/Project.toml
@@ -7,4 +7,4 @@ version = "0.1.0"
 julia = "1.6"
 
 [deps]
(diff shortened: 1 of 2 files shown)


pfitzseb:
Fun fact: You can just modify `Base.loaded_modules` after loading the first instances (basically pop the original entry and re-insert it with a modified UUID).

c42f:
Interesting. Is there a way to get this to be permanent when using PackageCompiler to create the sysimage?

pfitzseb:
Good question. I have no idea, tbh :)

c42f:
Ok let's go with this hack, then :)

PR #39, opened 2022-08-08
Docs: Add notes about concatenation AST forms

c42f:
@benchung I added some questions you've had about concatenation to the docs here


pfitzseb:
There's also the incredibly fun
```
julia> Meta.parse("[a b;;
       c d]")
:([a b c d])
```
case. Not sure if it makes sense covering that here though.

c42f:
yes indeed :) I guess I'll leave that out of this section which was mostly trying to cover the allowable AST forms.

Issue #33, opened 2022-06-24
Where should the lexer live?


c42f:
In https://github.com/JuliaLang/JuliaSyntax.jl/issues/31#issuecomment-1164175383, @pfitzseb said

> Btw, we really should think about upstreaming the Tokenize changes in this repo... Pretty sure the opsuffix changes for `&&`/`||` are implemented there.

We've chatted about this in various places and I mention it in the README. I'd like to resolve the double maintenance problem in some way, for sure :-)

But having modified Tokenize fairly extensively, I'm unsure whether the lexer should be versioned separately from the parser. Currently I see the lexer as serving the needs of parsing rather than something which is independent. Particularly because

* Lexing Julia correctly is impossible without keeping state. Worse, that state needs to be recursive for nested string interpolations. Other cases which need state or lookahead are prime (#25) and various contextual keywords like `outer`. It's possible to add state to the lexer itself, but that's annoyingly redundant. And the redundancy of state becomes much worse when you consider recovery from malformed string interpolations.
* JuliaSyntax can give you the disambiguated token stream in a flat format out of `ParseStream`. It's fairly lightweight, no need to opt into `Expr` (or other) tree building!
* Parsing+lexing is currently only about half as fast as pure lexing.

So with those in mind, I feel like we could just recommend people use the full parser for purposes we previously used Tokenize.jl for? And that more tightly integrating the tokenizer source into JuliaSyntax might be best.

(Somewhat of a side note — I've also wondered whether we could do an Automa.jl - based lexer if we wanted to delve more deeply into performance optimization. I suspect a generated lexer would be a lot faster if unicode decoding were folded into the state machine.)

For now, I'm content to port fixes back and forth as required.

What do people think? @pfitzseb @kristofferc ?

KristofferC:
I'm OK with soft deprecating Tokenize for full JuliaSyntax.

pfitzseb:
Same. The current situation (CSTParser being used, JuliaSyntax not so much) is a bit annoying, but the solution to that is to kill off CSTParser too :)

c42f:
I've started a big refactor to integrate Tokenize better into JuliaSyntax and delete a lot of the shim code I had to put in.

inkydragon:
> I suspect a generated lexer would be a lot faster if unicode decoding were folded into the state machine.

https://gershnik.github.io/2021/03/24/reverse-utf8-decoding.html
[...]

PR #40, opened 2022-08-10
Integrate Tokenize source more closely into JuliaSyntax

c42f:
As discussed in #33, I'm working to integrate my copy of Tokenize.jl into the larger JuliaSyntax codebase. This is the first batch of those changes:

* I've created a new `Kind` bits type in JuliaSyntax.jl and used the `@K_str` macro to consistently name these enumerants everywhere. This is in src/kind.jl.
* I've removed ~1000 lines of shim code related to mapping the old `Tokenize.Kind` to the `K_str` macro
* Used `@K_str` to name the enumerants in the tokenize source which makes things much more readable
* Renamed or moved a few predicates for consistency
* I've consolidated all of the Tokenize.jl source into `src/tokenize.jl` and `src/tokenize_utils.jl` and removed Tokenize project-level files which are no longer needed.  Also removed the `Tokens/Lexers` submodules which I think are unnecessary at this point.

### Attribution / license stuff

I've added some text to the main LICENSE.md to acknowledge the provenance of the Tokenize.jl code. Also, the full git history of Tokenize.jl is available in this repo.  In the process I removed a reference to the license text of the old JuliaParser.jl code, as any nontrivial use of that code seems long gone.


c42f:
@KristofferC, hope those license/attribution arrangements are good! Naturally you and other Tokenize authors are in the git history in any case.

src/kinds.jl
```diff
@@ -0,0 +1,1102 @@
...
+are unityped from the Julia compiler's point of view, for efficiency.
+
+Naming rules:
+* Kinds which correspond to exactly one textural form are represented with that
+  text. This includes keywords like K"for" and operators like K"*".
+* Kinds which represent many textural forms have UpperCamelCase names. This
+  includes kinds like K"Identifier" and K"Comment".
+* Kinds which exist merely as delimiters are all uppercase
+"""
+primitive type Kind 16 end
```
c42f, line 918: Mostly this PR is code movement, but the new `Kind` type here is new code.

LICENSE.md
```diff
@@ -1,4 +1,27 @@
-The Tokenize.jl package is licensed under the MIT "Expat" License:
+The JuliaSyntax.jl package is licensed under the MIT "Expat" License:
+
+> Copyright (c) 2021 Julia Computing and contributors
```
davidanthoff, line 3: Wouldn't it be easier to just add @KristofferC here on this line? I generally think it is good if LICENSE files just have the template language and no custom additions, makes it easier for users. Plus, in this case I doubt that legally it makes any difference.

I _do_ think it would be nice to ack @KristofferC explicitly, not for legal reasons but just for history, but maybe the README is a better place for that?

c42f: Sure it'd be cleaner to consolidate.  But also I don't care too much about how clean the license file is. Whatever makes people happy.

Tokenize.jl is currently linked to from the README.

KristofferC: I personally don't care too much about what is in the LICENSE.

c42f: In that case, how about this: I reverted to the original license text for JuliaSyntax.jl. That was somewhat author-neutral and It's all MIT anyway.

git has the real history.

c42f: Hmm I didn't feel quite right about this so I've gone back to including the Tokenize license how I had it. Will merge shortly; we can always simplify later if desired.

PR #46, opened 2022-08-15
Reduce invalidation/recompliation for parser usage in sysimage

c42f:
* Freeze world age by default in enable_in_core!()
* Use a `Ref` to hide the invocation of the main hook function from the compiler

I don't understand why the use of the `Ref` (the second commit) seems necessary here; I would have thought that `invoke_in_world` should have done the job. But when I tested on Julia 1.7.3 without the `Ref`, loading the `TerminalLoggers` package caused significant latency on the next parse time.

Presumably the `Ref` breaks some extra backedges, though I'm not sure what they are or how best to debug the situation. (Perhaps `SnoopCompile`? I'd need to learn how to use that tooling.)

### DIFF
--- a/src/hooks.jl
+++ b/src/hooks.jl
@@ -1,14 +1,22 @@
-# World age which the parser will be invoked in.
-# Set to typemax(UInt) to invoke in the world of the caller.
-const _parser_world_age = Ref{UInt}(typemax(UInt))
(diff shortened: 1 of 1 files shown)


c42f:
This also seems to work in 1.9, so let's go with it for now.

Issue #51, opened 2022-08-17
Internal error in parser when failing to parse a Float64


KristofferC:
```jl
julia> 0x1p
┌ Error: JuliaSyntax parser failed — falling back to flisp!
│   exception =
│    ArgumentError: cannot parse "0x1p" as Float64
│    Stacktrace:
│      [1] _parse_failure(T::Type, s::String, startpos::Int64, endpos::Int64) (repeats 2 times)
│        @ Base ./parse.jl:373
│      [2] #tryparse_internal#494
│        @ ./parse.jl:369 [inlined]
│      [3] tryparse_internal
│        @ ./parse.jl:366 [inlined]
│      [4] #parse#495
│        @ ./parse.jl:379 [inlined]
│      [5] parse
│        @ ./parse.jl:379 [inlined]
│      [6] julia_string_to_number(str::SubString{String}, kind::JuliaSyntax.Kind)
│        @ JuliaSyntax ~/julia/usr/share/julia/stdlib/v1.9/JuliaSyntax/src/value_parsing.jl:28
│      [7] JuliaSyntax.SyntaxNode(source::JuliaSyntax.SourceFile, raw::JuliaSyntax.GreenNode{JuliaSyntax.SyntaxHead}, position::UInt32)
│        @ JuliaSyntax ~/julia/usr/share/julia/stdlib/v1.9/JuliaSyntax/src/syntax_tree.jl:33
│      [8] JuliaSyntax.SyntaxNode(source::JuliaSyntax.SourceFile, raw::JuliaSyntax.GreenNode{JuliaSyntax.SyntaxHead}, position::UInt32)
│        @ JuliaSyntax ~/julia/usr/share/julia/stdlib/v1.9/JuliaSyntax/src/syntax_tree.jl:87
│      [9] build_tree(::Type{JuliaSyntax.SyntaxNode}, stream::JuliaSyntax.ParseStream; filename::String, kws::Base.Pairs{Symbol, JuliaSyntax.Kind, Tuple{Symbol}, NamedTuple{(:wrap_toplevel_as_kind,), Tuple{JuliaSyntax.Kind}}})
│        @ JuliaSyntax ~/julia/usr/share/julia/stdlib/v1.9/JuliaSyntax/src/syntax_tree.jl:187
```

I believe this should emit an error instead of throwing inside the parser.

c42f:
Yes, the parser should never throw. Same for the tree conversion code, I think, which is where this error is.

PR #52, opened 2022-08-17
require a number after 'p' for hex floats to lex as a float

KristofferC:
Should fix https://github.com/JuliaLang/JuliaSyntax.jl/issues/51.


c42f (approved):
looks good to me!

Issue #59, opened 2022-08-22
Incorrect parsing of array literal with newline before comma


oscardssmith:
On Julia master,
```
[ []
, [] ]
```
parses correctly as a `Vector{Vector{Any}}`, but with JuliaSyntax, we get the `ParseError`
```
Error: Expected `]`
@ REPL[11]:2:3
[ []
  ,[] ]
```
This is a fairly major bug since this prevents loading CpuId which depends on this syntax.

c42f:
> fairly major bug

Definitely a problematic bug! Whether it's a more major one than the other known bugs I don't know :sweat_smile:

PR #60, opened 2022-08-22
Fix vect parsing with newline before comma

pfitzseb:
Fixes https://github.com/JuliaLang/JuliaSyntax.jl/issues/59.

### DIFF
--- a/src/parser.jl
+++ b/src/parser.jl
@@ -2798,7 +2798,7 @@ function parse_cat(ps::ParseState, closer, end_is_symbol)
     if k == K"," || (is_closing_token(ps, k) && k != K";")
         if k == K","
             # [x,]  ==>  (vect x)
-            bump(ps, TRIVIA_FLAG)
+            bump(ps, TRIVIA_FLAG; skip_newlines = true)
         end
         # [x]      ==>  (vect x)
         # [x \n ]  ==>  (vect x)
@@ -2957,9 +2957,6 @@ function parse_brackets(after_parse::Function,
             num_semis += 1
             bump(ps, TRIVIA_FLAG)
             bump_trivia(ps)
-        elseif k == K","
-            had_commas = true
-            bump(ps, TRIVIA_FLAG)
         elseif is_closing_token(ps, k)
             # Error; handled below in bump_closing_token
             break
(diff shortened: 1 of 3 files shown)


oscardssmith:
Looks good. Merge?

pfitzseb:
I'll merge this tomorrow morning, unless Chis chimes in and says this is the wrong fix :)

The test is currently a bit awkward, because SyntaxNode/Expr conversion can't handle unexpected `,`s. Would like to fix that.

test/parser.jl
```diff
@@ -779,6 +782,10 @@ broken_tests = [
     ]
 ]
 
+@testset "Invalid syntax" begin
+    @test !isempty(JuliaSyntax.parse(JuliaSyntax.GreenNode, """[ [],,[] ]""")[2])
```
c42f, line 786: Let's make a more precise test case out of this and add it to the vector of `tests`.

The parser should emit some `error` node and we can test that.

src/parser.jl
```diff
@@ -2798,7 +2798,7 @@ function parse_cat(ps::ParseState, closer, end_is_symbol)
     if k == K"," || (is_closing_token(ps, k) && k != K";")
         if k == K","
             # [x,]  ==>  (vect x)
-            bump(ps, TRIVIA_FLAG)
+            bump(ps, TRIVIA_FLAG; skip_newlines = true)
```
c42f, line 2801: Cool looks good. I wonder how many more cases like this there are elsewhere in the parser ... I've found a few before ...

c42f (commented):
Looks good thanks.  Let's add such error cases to the tests as s-expressions along with the happy cases.

c42f:
> The test is currently a bit awkward, because SyntaxNode/Expr conversion can't handle unexpected

Ah yeah I didn't see this comment before I wrote the review.  Precisely :)

pfitzseb:
Tests look a lot nicer now. We're now just emitting an `ErrorVal` when encountering unknown node kinds, which may be a bad idea, but at least allows us to recover from the corrupted tree we produce further down :P

pfitzseb:
This good to go, or do you think emitting `ErrorVal`s instead of actually erroring is going to be an issue further down the line?

c42f (approved):
The FIXME with `ErrorVal` makes me a bit uncomfortable (does seem wrong). But this is still an improvement so I think we can merge and fix that later.

PR #68, opened 2022-08-23
Add check_all_packages script

pfitzseb:
I ran this with https://github.com/JuliaLang/JuliaSyntax.jl/pull/60 and ended up with
```
152900 files parsed
  37 failures
  0.82ms per file, 126s in total
```
Issues are open for most of those failures.

### DIFF
--- a/tools/check_all_packages.jl
+++ b/tools/check_all_packages.jl
@@ -1,0 +1,64 @@
+# hacky script to parse all Julia files in all packages in General
+# to Exprs and report errors
+#
+# Run this after registry_download.jl (so the pkgs directory is populated).
+
+using JuliaSyntax, Logging
+
+logio = open(joinpath(@__DIR__, "logs.txt"), "w")
+logger = Logging.ConsoleLogger(logio)
+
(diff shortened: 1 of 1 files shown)


tools/check_all_packages.jl
```diff
@@ -0,0 +1,64 @@
...
+logio = open(joinpath(@__DIR__, "logs.txt"), "w")
+logger = Logging.ConsoleLogger(logio)
+
+pkgspath = joinpath(@__DIR__, "pkgs")
+
+parallel = 50
+exceptions = []
+Logging.with_logger(logger) do
+    for tars in Iterators.partition(readdir(pkgspath), parallel)
+        @sync for tar in tars
```
c42f, line 17: Is this partition essentially doing the same as `asyncmap()` but manually?

c42f (approved):
Looks good to me

c42f:
PS, this is great btw, thanks for running with what I started with registry_download.jl but didn't get around to finishing!

tools/check_all_packages.jl
pfitzseb: Yup, looks like it :)

PR #71, opened 2022-08-24
generate `:incomplete` expression

### LINKED ISSUE #42: When using in REPL, `JuliaSyntax.jl` does not seem to work well with multiline parsing?
When I want to type

```julia
map(1:10) do x
    2x
end
```

after I type `map(1:10) do x` and hit return, I got

```julia
julia> map(1:10) do
ERROR: ParseError:
Error: premature end of input
@ REPL[11]:2:1
map(1:10) do
Error: premature end of input
@ REPL[11]:2:1
map(1:10) do
Error: Expected `end` but got unexpected tokens
@ REPL[11]:2:1
map(1:10) do
```

Or if I want to type

```julia
 w = u +
     v +
     1
```

I got

```julia
julia> w = u +
ERROR: ParseError:
Error: premature end of input
@ REPL[14]:2:1
w = u +
```

But I can do them with copy-paste:

image

Can we make the REPL-mode tolerate multiline code a little bit more?

aviatesk:
This commit setups a specific diagnostic level `:incomplete` to distinguish
incomplete cases from the other general error cases.
Does this approach sound reasonable?

fix #42

### DIFF
--- a/src/diagnostics.jl
+++ b/src/diagnostics.jl
@@ -29,17 +29,22 @@ struct Diagnostic
     message::String
 end
 
-function Diagnostic(first_byte, last_byte; error=nothing, warning=nothing)
-    message = !isnothing(error)   ? error :
-              !isnothing(warning) ? warning :
+function Diagnostic(first_byte, last_byte; error=nothing, warning=nothing, incomplete=nothing)
+    message = !isnothing(error)      ? error :
+              !isnothing(warning)    ? warning :
+              !isnothing(incomplete) ? incomplete :
               Base.error("No message in diagnostic")
-    level = !isnothing(error) ? :error : :warning
+    level = !isnothing(error) ? :error : !isnothing(incomplete) ? :incomplete : :warning
     Diagnostic(first_byte, last_byte, level, message)
 end
 
-first_byte(d::Diagnostic) = d.first_byte
-last_byte(d::Diagnostic)  = d.last_byte
-is_error(d::Diagnostic)   = d.level == :error
+first_byte(d::Diagnostic)    = d.first_byte
+last_byte(d::Diagnostic)     = d.last_byte
+is_error(d::Diagnostic)      = d.level == :error
+is_incomplete(d::Diagnostic) = d.level == :incomplete
+
+any_error(ds::AbstractVector{Diagnostic}) = any(is_error, ds)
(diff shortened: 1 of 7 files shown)


c42f:
I had a different plan for this. My thought was that we'd detect this during post-processing: if the first diagnostic is an error and  touches the last byte of the input, that implies an incomplete expression.  Then the parser doesn't need to know about this at all.  In a way I like that approach, because I feel like `incomplete` and `error` are overlapping categories.

I don't have time to finish my version today, though I could push my WIP :)

c42f:
I should say I'm not sure which approach works best in practice. If we go with the idea here that may be fine too. I think maybe `incomplete` is better as a flag in that case.

In any case, to have this all work well with the REPL, we need `Base.incomplete_tag` to work?  Here's a quick paste of some test cases I scavenged from Base:


```julia
    @testset "Expr(:incomplete)" begin
        for (str, tag) in [
               ""                  => :none
               "\""                => :string
               "#="                => :comment
               "'"                 => :char
               "`"                 => :cmd
               "begin;"            => :block
               "quote;"            => :block
               "let;"              => :block
               "for i=1;"          => :block
               "function f();"     => :block
               "f() do x;"         => :block
               "module X;"         => :block
               "mutable struct X;" => :block
               "struct X;"         => :block
               "("                 => :other
               "["                 => :other
               "begin"             => :other
               "quote"             => :other
               "let"               => :other
               "for"               => :other
               "function"          => :other
               "f() do"            => :other
               "module"            => :other
               "mutable struct"    => :other
               "struct"            => :other
            ]
            @test Base.incomplete_tag(Meta.parse(str, raise=false)) == tag
        end
        # @test Meta.isexpr(Meta.parse("1 == 2 ?"), :incomplete)
        # @test Meta.isexpr(Meta.parse("1 == 2 ? 3 :"), :incomplete)
        # @test Meta.isexpr(Meta.parse("1,"), :incomplete)
        # @test Meta.isexpr(Meta.parse("1, "), :incomplete)
        # @test Meta.isexpr(Meta.parse("1,\n"), :incomplete)
        # @test Meta.isexpr(Meta.parse("1, \n"), :incomplete)
        # @test Base.incomplete_tag(Meta.parse("\"\$foo", raise=false)) === :string
        # @test Meta.isexpr(Meta.parse("'a"), :incomplete)
    end

```

Issue #66, opened 2022-08-23
`dim` can overflow in `parse_array`


pfitzseb:
E.g. when parsing https://github.com/OpenMendel/MendelImpute.jl/blob/v1.2.3/test/run.jl. That file's obviously invalid, but we should presumably handle this case a bit more nicely. `Meta.parse` doesn't error here.

```
┌ Error: parsing failed for /home/pfitzseb/juliasyntaxtest/pkgs/MendelImpute_1.2.3/test/run.jl
│   ex =
│    Numeric flags unable to hold large integer -9223372036854775808
│    Stacktrace:
│      [1] error(s::String)
│        @ Base ./error.jl:35
│      [2] set_numeric_flags
│        @ ~/.julia/packages/JuliaSyntax/OawBx/src/parse_stream.jl:34 [inlined]
│      [3] parse_array(ps::JuliaSyntax.ParseState, mark::JuliaSyntax.ParseStreamPosition, closer::JuliaSyntax.Kind, end_is_symbol::Bool)
│        @ JuliaSyntax ~/.julia/packages/JuliaSyntax/OawBx/src/parser.jl:2634
│      [4] parse_cat(ps::JuliaSyntax.ParseState, closer::JuliaSyntax.Kind, end_is_symbol::Bool)
│        @ JuliaSyntax ~/.julia/packages/JuliaSyntax/OawBx/src/parser.jl:2814
│      [5] parse_call_chain(ps::JuliaSyntax.ParseState, mark::JuliaSyntax.ParseStreamPosition, is_macrocall::Bool)
│        @ JuliaSyntax ~/.julia/packages/JuliaSyntax/OawBx/src/parser.jl:1469
│      [6] parse_call_chain
│        @ ~/.julia/packages/JuliaSyntax/OawBx/src/parser.jl:1383 [inlined]
│      [7] parse_call(ps::JuliaSyntax.ParseState)
│        @ JuliaSyntax ~/.julia/packages/JuliaSyntax/OawBx/src/parser.jl:1311
│      [8] parse_factor(ps::JuliaSyntax.ParseState)
│        @ JuliaSyntax ~/.julia/packages/JuliaSyntax/OawBx/src/parser.jl:1258
│      [9] parse_unary(ps::JuliaSyntax.ParseState)
│        @ JuliaSyntax ~/.julia/packages/JuliaSyntax/OawBx/src/parser.jl:1101
[...]

PR #72, opened 2022-08-24
Fix early-abort handling in array_parse

pfitzseb:
Fixes https://github.com/JuliaLang/JuliaSyntax.jl/issues/66.

### DIFF
--- a/src/parser.jl
+++ b/src/parser.jl
@@ -2617,6 +2617,9 @@ function parse_array(ps::ParseState, mark, closer, end_is_symbol)
     # For an excellent overview of Pratt parsing, see
     # https://matklad.github.io/2020/04/13/simple-but-powerful-pratt-parsing.html
     (dim, binding_power) = parse_array_separator(ps, array_order)
+    if binding_power == typemin(Int)
+        return (K"hcat", EMPTY_FLAGS)
+    end
     while true
         (next_dim, next_bp) = parse_array_inner(ps, binding_power, array_order)
         if next_bp == typemin(Int)
--- a/src/parser.jl
+++ b/src/parser.jl
@@ -2618,6 +2618,7 @@ function parse_array(ps::ParseState, mark, closer, end_is_symbol)
     # https://matklad.github.io/2020/04/13/simple-but-powerful-pratt-parsing.html
     (dim, binding_power) = parse_array_separator(ps, array_order)
     if binding_power == typemin(Int)
+        # [x@y  ==>  (hcat x (error-t ✘ y))
         bump_closing_token(ps, closer)
         return (K"hcat", EMPTY_FLAGS)
     end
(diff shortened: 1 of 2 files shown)


src/parser.jl
```diff
@@ -2617,6 +2617,9 @@ function parse_array(ps::ParseState, mark, closer, end_is_symbol)
     # For an excellent overview of Pratt parsing, see
     # https://matklad.github.io/2020/04/13/simple-but-powerful-pratt-parsing.html
     (dim, binding_power) = parse_array_separator(ps, array_order)
+    if binding_power == typemin(Int)
+        return (K"hcat", EMPTY_FLAGS)
```
c42f, line 2621: Let's consume `]` if it's present.

```suggestion
        bump_closing_token(ps, closer)
        return (K"hcat", EMPTY_FLAGS)
```

c42f (commented):
Let's ensure `parse_array` looks for the closer for consistency.  Still doesn't give the best error, but this does seem like an obscure edge case.

BTW, there's a tool `itest_parse` for interactive development showing the various trees and the flisp parser output

```
julia> include("test/test_utils.jl")

julia> itest_parse(JuliaSyntax.parse_expr, "[a@b]")
# Code:
[a@b]

# Green tree:
     1:5      │[hcat]
     1:1      │  [                          "["
     2:2      │  Identifier             ✔   "a"
     3:4      │  [error]                 ✘
     3:3      │    @                    ✔   "@"
     4:4      │    Identifier           ✔   "b"
     5:5      │  ]                          "]"
Error: Expected `]`
@ line 1:3
[a@b]

# SyntaxNode:
(hcat a (error-t ✘ b))
```

c42f (approved):

src/parser.jl
```diff
@@ -2617,6 +2617,10 @@ function parse_array(ps::ParseState, mark, closer, end_is_symbol)
     # For an excellent overview of Pratt parsing, see
     # https://matklad.github.io/2020/04/13/simple-but-powerful-pratt-parsing.html
     (dim, binding_power) = parse_array_separator(ps, array_order)
+    if binding_power == typemin(Int)
+        bump_closing_token(ps, closer)
```
c42f, line 2621: Oh let's also add the case that can lead to this branch inline. It makes reading the code later a *lot* easier

```suggestion
        # [x@y  ==>  (hcat x (error-t ✘ y))
        bump_closing_token(ps, closer)
```

c42f: My plan has kind of been to extract these inline comments automatically and use them as test cases ... haven't got around to that yet though

pfitzseb: Yeah, I saw that comment in the tests. Will be a bit annoying to keep the inline annotations in sync with the tests until that's done.

Issue #61, opened 2022-08-22
Error message in plain text discards visual location information


ForceBru:
When I get a syntax error, the offending characters are highlighted:

![syntaxError](https://user-images.githubusercontent.com/8695383/185992651-d6fea2fe-9e8f-47e0-9857-6b85fb7705a5.png)

In fact, they're... "lowlighted", so that the offending character is _dimmer_ than other code (I can barely see the `8` there), but it's still (de)emphasized.

When I copy this error message and paste it as plain text, all formatting is lost, which is expected:

```julia
julia> (2+5+8
ERROR: ParseError:
Error: Expected `)` but got unexpected tokens
@ REPL[17]:2:1
(2+5+8
```

However, nothing _visually points_ at the location of the error anymore: the `8` is no different from other code. Sure, the location is indicated by `REPL[17]:2:1` (Does it mean "second line, first character"? Looks more like a "1st line, 6th character" to me...), but the visual is lost. Implementations of other programming languages rely on actual characters (not formatting) to indicate the position of the error.

----------

Python draws a "pointer" to the error location, so a plain-text error message still tells me where the error is:

```python
>>> (2,4,;)
  File "<stdin>", line 1
    (2,4,;)
         ^
SyntaxError: invalid syntax
```

Clang does this too:

```
$ cat syntax_error.c 
int main() {
    0
}
$ clang syntax_error.c 
syntax_error.c:2:6: error: expected ';' after expression
    0
     ^
     ;
syntax_error.c:2:5: warning: expression result unused [-Wunused-value]
    0
    ^
1 warning and 1 error generated.
```

Rust's error messages [draw around the offending code all the time](https://rustc-dev-guide.rust-lang.org/diagnostics.html#diagnostic-structure), which is extremely helpful.

------

It would be nice to have error messages one could copy & paste without loss of information, especially _visual_ indication of where the error is.

- Julia 1.8.0
- JuliaSyntax.jl 0.1.0

c42f:
> In fact, they're... "lowlighted

This shouldn't happen and seems like a bug in the highlighting code, or more likely a problem that the code currently assumes high quality color in the terminal emulator.

Perhaps you've only got 16 or 256 terminal background colors. Which terminal are you using?

> "pointer" to the error location

I'd definitely like a text-only mode as an option. But I think this would be worse in two key ways:
* It disturbs the visual shape of the code much more than a simple highlighting scheme
[...]

PR #74, opened 2022-08-24
change highlight color from an RGB value to `Base.error_color()`

KristofferC:
RGB colors have very spotty support in terminals so probably better to use something more standard. `Base.error_color()` is by default `:light_red` but can also be customized with an env var.

Fixes https://github.com/JuliaLang/JuliaSyntax.jl/issues/61

### DIFF
--- a/src/diagnostics.jl
+++ b/src/diagnostics.jl
@@ -79,7 +79,7 @@ function show_diagnostic(io::IO, diagnostic::Diagnostic, source::SourceFile)
     a,b = source_line_range(source, p, context_lines_before=2, context_lines_after=1)
     c,d = source_line_range(source, q, context_lines_before=1, context_lines_after=2)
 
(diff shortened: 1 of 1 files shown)


c42f:
Yeah `_printstyled()` is my own machinery because `Base` isn't very flexible, so the base color codes won't work here.

> RGB colors have very spotty support in terminals so probably better to use something more well-supported

I used RGB to satisfy my own desire for things to be pretty and to have sane mappings to color codes. But for general use we'll need something widely supported :+1:

Unfortunately `Base.printstyled` can't do background colors I think? And it can't do `reverse` until Julia 1.7.  So we probably still need a little of our own machinery here.

So we "just" need to figure out what's generally supported and do that.  Can we use 255 colors for example, or do we need to limit ourselves to 16?  Is reverse color better supported than background colors?

PR #77, opened 2022-08-25
Better "unexpected kw" handling in ternary parsing

pfitzseb:


src/parser.jl
```diff
@@ -642,6 +642,9 @@ function parse_cond(ps::ParseState)
         # a ? b :c  ==>  (if a [ ] [?] [ ] b [ ] [:] (error-t) c)
         bump_invisible(ps, K"error", TRIVIA_FLAG, error="space required after `:` in `?` expression")
     end
+
+    # FIXME: This is a very specific case. Error recovery should be handled mor
```
KristofferC, line 646:
```suggestion
    # FIXME: This is a very specific case. Error recovery should be handled more
```

src/parser.jl
```diff
@@ -637,11 +637,25 @@ function parse_cond(ps::ParseState)
...
+    t = peek_token(ps; skip_newlines = true)
     if !preceding_whitespace(t)
         # a ? b :c  ==>  (if a [ ] [?] [ ] b [ ] [:] (error-t) c)
-        bump_invisible(ps, K"error", TRIVIA_FLAG,
-                       error="space required after `:` in `?` expression")
+        bump_invisible(ps, K"error", TRIVIA_FLAG, error="space required after `:` in `?` expression")
+    end
+
+    # FIXME: This is a very specific case. Error recovery should be handled more
+    # generally elsewhere.
```
c42f, line 647: aha agreed! This is why I've been wondering what to do about this PR.

tools/check_all_packages.jl
```diff
@@ -5,6 +5,23 @@
 
 using JuliaSyntax, Logging
 
+# like Meta.parseall, but throws
+function parseall(str)
```
c42f, line 9: maybe call this something other than `parseall`, given that we already have `JuliaSyntax.parseall`? It may just be confusing.

c42f (approved):
I think this is fine, as far as it goes.

I've been procrastinating about doing this review because I want to do error reporting more like I've vaguely outlined in #93.

But to motivate a general system like #93 we do need a bunch of specific use cases like this, #92 and all the other specific error messages which are hardcoded in various places.

So as long as error recovery is clearly marked and doesn't cause large scale disruption of the parser I think we can add it piecemeal for now.

PR #80, opened 2022-08-30
Allow newline in range expr after : in parens

pfitzseb:


src/parser.jl
```diff
@@ -822,9 +822,10 @@ function parse_range(ps::ParseState)
                 emit_diagnostic(ps, error="found unexpected closing token")
                 return
             end
-            if had_newline
+            if had_newline && !ps.whitespace_newline
```
c42f, line 825: Perhaps we should use `had_newline = peek(ps) == K"NewlineWs"` further up instead?  That should take `whitespace_newline` into account automatically.

I'm not sure why I have `skip_newlines=false` in there currently.

pfitzseb: I'm not sure either, but assumed it was for a reason. Tests pass without `skip_newlines=false`, so let's go with that instead.

c42f (approved):

Issue #79, opened 2022-08-30
Failure to parse blocks in comprehension if condition


pfitzseb:
```
julia> JuliaSyntax.parse(Expr, """
       Any[foo(i) 
           for i in x if begin
               true
           end
       ]
       """)
(:($(Expr(:toplevel, :(#= line 1 =#), :($(Expr(:typed_comprehension, :Any, :((foo(i) for i = x if begin)), :($(Expr(:error, true, :($(Expr(:end)))))))))))), JuliaSyntax.Diagnostic[JuliaSyntax.Diagnostic(45, 56, :error, "Expected `]`")], 60)

julia> Meta.parse("""
       Any[foo(i) 
           for i in x if begin
               true
           end
       ]
       """)
:(Any[foo(i) for i = x if begin
          #= none:3 =#
          true
      end])

julia> Meta.parse("""
       Any[foo(i) for i in x if begin
               true
           end
       ]
       """)
ERROR: Base.Meta.ParseError("expected \"]\"")
```
Note that `Meta.parse` is sensitive to the newline before `for`, so it's possible we should treat this as a bug in the reference parser.

Original code can be found [here](https://github.com/poldavezac/BokehServer.jl/blob/db9ba4e3dab33a968ac4108e8bb8305e6d9fc493/src/core/protocol/patchdoc.jl#L13-L26).

PR #81, opened 2022-08-30
Correctly handle begin as kw in typed comprehension

pfitzseb:
Fixes https://github.com/JuliaLang/JuliaSyntax.jl/issues/79.

### DIFF
--- a/src/parser.jl
+++ b/src/parser.jl
@@ -2525,6 +2525,7 @@ end
 # then reconstructing the nested flattens and generators when converting to Expr.
 #
 # [x for a = as for b = bs if cond1 for c = cs if cond2] ==> (comprehension (flatten x (= a as) (filter (= b bs) cond1) (filter (= c cs) cond2)))
+# [x for a = as if begin cond2 end]  =>  (comprehension (generator x (filter (= a as) (block cond2))))
 #
 # flisp: parse-generator
 function parse_generator(ps::ParseState, mark, flatten=false)
@@ -2561,7 +2562,8 @@ end
 # flisp: parse-comprehension
 function parse_comprehension(ps::ParseState, mark, closer)
     ps = ParseState(ps, whitespace_newline=true,
-                    space_sensitive=false)
+                    space_sensitive=false,
+                    end_symbol=false)
     parse_generator(ps, mark)
     bump_closing_token(ps, closer)
     return (K"comprehension", EMPTY_FLAGS)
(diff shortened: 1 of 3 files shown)


src/parser.jl
```diff
@@ -2561,7 +2562,8 @@ end
 # flisp: parse-comprehension
 function parse_comprehension(ps::ParseState, mark, closer)
     ps = ParseState(ps, whitespace_newline=true,
-                    space_sensitive=false)
+                    space_sensitive=false,
+                    end_symbol=false)
```
c42f, line 2566: The thing that confuses me about this is why the flisp parser doesn't seem to have this state change within `parse-comprehension`. So I wonder whether this is the right fix, or whether we're still missing something here.

pfitzseb: Definitely possible. I haven't dug into the scheme parser much (because I find it borderline unreadable :P), but this seems correct to me regardless. Note that Julia currently can't handle this syntax in the general case:
```
julia> Meta.parse("""
       Any[foo(i) for i in x if begin
               true
           end
       ]
       """)
ERROR: Base.Meta.ParseError("expected \"]\"")
```
The newline before `for` is required for this to work.

c42f: Yeah that's quite weird.  Certainly seems like a bug in `Meta.parse()`.

I'm keeping a list of such bugs in the README - can you add it here? https://github.com/JuliaLang/JuliaSyntax.jl#flisp-parser-bugs

pfitzseb: Done.

c42f:
> why the flisp parser doesn't seem to have this state change within parse-comprehension

So I think I understood what's going on here.

The only place the flisp parser sets `end-symbol` is when parsing "array indexing" (ie, `Expr(:ref)`) which shares a syntactic prefix up until the `for` is encountered:

https://github.com/JuliaLang/julia/blob/73f9542055d7160c47659269c05d801d1d8fd009/src/julia-parser.scm#L1223

So now with this change, we've got somewhat of an inconsistency that the part before the `for` uses `begin` and `end` as identifiers, but the part after the `for` has `begin ... end` as a block.

c42f:
> somewhat of an inconsistency

Ah, though I now see this isn't worse than the flisp parser, which is already inconsistent in the same way for `typed_hcat`:

```
julia> dump(:(a[begin begin b end]))
Expr
  head: Symbol typed_hcat
  args: Array{Any}((3,))
    1: Symbol a
    2: Symbol begin
    3: Expr
      head: Symbol block
      args: Array{Any}((2,))
        1: LineNumberNode
          line: Int64 1
          file: Symbol REPL[5]
        2: Symbol b
```

PR #85, opened 2022-08-31
Use `K"?"` for head of ternary conditional

c42f:
This allows `a ? b : c` syntax to be clearly distinguished from normal
`if a b else c end` without reparsing syntax trivia.

For compatibility we convert to `:if` for normal `Expr`


c42f:
(This change brought to mind in a discussion with @JeffBezanson today)

pfitzseb (approved):

PR #86, opened 2022-08-31
Handle error nodes in Expr construction

pfitzseb:
This makes `parse(Expr, ...)` succeed even when error nodes were emitted previously. 

I'm not sure we even want this, but I feel like there's some value to never getting an exception when parsing arbitrary invalid code.


c42f:
Superseded by #224. (Though also, I think we need a whole new way of communicating diagnostics to the runtime outside of `Expr`.)

Issue #88, opened 2022-09-01
Reworking some Julia AST forms


c42f:
With JuliaSyntax, we've got our own green tree (`GreenNode`) and AST (`SyntaxNode`) which often differ from `Expr`, due to the requirement that children are strictly in source order.  Some current differences are described in https://github.com/JuliaLang/JuliaSyntax.jl#tree-differences-between-greennode-and-expr. Given that we've been forced to diverge we might as well make the most of this and reconsider some aspects of `Expr` for two reasons:

* To give extra textural information to users of green tree (eg, presence of parentheses), without the need to recompute this information by inspecting the trivia. (These nodes will probably be elided during AST conversion?)
* To give extra information to AST users like macro writers (eg, conditional ternary is the same as `:if` for `Expr` users)

List of possible changes

* [x] Insertion of extra blocks inconsistent in various places, sometimes due to the need of `Expr` to hold `LineNumberNode`.  Avoid this!
    * [x] `let` argument lists (#126)
    * [x] short form functions have a block for the body line number
    * [x] `elseif` conditional has a block for the line number
    * [x] anon function `->` has a block only for the body line number
* [x] Containers for string-like constructs - these are very useful for tooling which works with the tokens of the source text. Also separating the delimiters from the text they delimit removes a whole class of tokenization errors and lets the parser deal with them.
    * [x] string (always use K"string" to wrap strings, even when they only contain a single `K"String"` token. See #94)
    * [x] char (add a new `K"char"` for containing characters along with their delimiters?). See #121 
    * [x] backticks (use `K"`"` or add `K"backticks"`?)
    * [x] `var""` syntax (use `K"var"`? as the head?) (#127)
* [x] Faithful representation of the source text / avoid premature lowering
    * [x] `K"macrocall"` - allow users to easily distinguish macrocalls with parentheses from those without them (#218)
    * [x] parentheses  (`K"parens"`)
    * [x] Ternary conditional (#85)
    * [x] `global const` is normalized to `const global` in the parser. Should be done in `Expr` conversion (#130)
    * [x] The AST for `do` seems partially lowered. Should it be flatter like `f(x) do y ; body end` being `(do (call f) (tuple y) (block body))`? (#98)
    * [x] Lower `@.` to `@__dot__` later, not inside parser
    * [x] Lower docstrings to `Core.@doc` later, not inside parser (#217)
* [x] Improvements for AST inconsistencies
[...]

PR #94, opened 2022-09-06
Always use string nodes wrapping string chunks

c42f:
Ensures that strings are always encapsulated within a
`string` internal node of the parse tree, giving a place to include the
delimiters and unifying interpolated strings / strings with internal
whitespace with plain string literals.

~~Also, wrap `K"$"` nodes around interpolations in strings~~ - update: removed this part

That is, parse
  `"a $x b"`
as
  `(string "a " ($ x) " b")`

This makes the presence of interpolation syntax explicit, just as it is
for interpolation into expression literals.

There's two advantages of this

* We don't need a hack to distinguish "$("str")" from "str". This seems
  necessary now that all strings are wrapped in K"string" nodes, as the
  older hack doesn't work (hmm on second thoughts ... I think?)
* The `$` node gives a place to attach delimiter trivia, which makes it
  easy to distinguish `"$(x)"` from `"$x"`. (Though with a `K"parens"`
  node coming soon maybe that's not necessary?)

I'm not quite 100% sure we want this second part with the interpolation nodes ... might need to think about it a little more.

Part of #88

### DIFF
--- a/README.md
+++ b/README.md
@@ -450,6 +450,17 @@ The syntax `a ? b : c` is the same as `if a b else c` in `Expr` so macros can't
 distinguish these cases. Instead, we use a distinct expression head `K"?"` and
 lower to `Expr(:if)` during `Expr` conversion.
 
+### String nodes always wrapped in `K"string"` or `K"cmdstring"`
+
+All strings are surrounded by a node of kind `K"string"`, even non-interpolated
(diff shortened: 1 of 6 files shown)


c42f:
> I'm not quite 100% sure we want this second part with the interpolation nodes

Thinking about this more ... this introduces extra depth to the tree without really bringing much benefit. So I think I'll remove this part.

PR #96, opened 2022-09-09
Remove many unnecessary allocations

c42f:
* Eliminate the use of Ref to communicate values out of the anonymous function passed to parse_brackets(). It seems this forces unnecessary allocation of the Ref captured by the closure.
* Force specialization on `down` in `parse_assignment_with_initial_ex` to avoid an unnecessary generic dispatch there.

This should probably bring down the total number of allocations during parsing to around O(log(N) + M) in the length of the file N and depth, M, of the AST.

### DIFF
--- a/src/parser.jl
+++ b/src/parser.jl
@@ -540,7 +540,7 @@ function parse_assignment(ps::ParseState, down, equals_is_kw::Bool)
     parse_assignment_with_initial_ex(ps, mark, down, equals_is_kw)
 end
 
-function parse_assignment_with_initial_ex(ps::ParseState, mark, down, equals_is_kw::Bool)
+function parse_assignment_with_initial_ex(ps::ParseState, mark, down::T, equals_is_kw::Bool) where {T} # where => specialize on `down`
     t = peek_token(ps)
     k = kind(t)
     if !is_prec_assignment(k)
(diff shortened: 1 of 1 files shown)


c42f:
Thanks @nhdaly and @vilterp for a very useful tutorial on the new allocation profiler :-) Tracking these allocations down without it would have been quite hard.

NHDaly:
Amazing! Awesome to see that! :)

I just did one of these, too, over here, using the tool too:
https://github.com/JuliaConcurrent/MultiThreadedCaches.jl/pull/10

Love to see it :)

src/parser.jl
```diff
@@ -1169,19 +1169,18 @@ function parse_unary_call(ps::ParseState)
...
-                    eq_is_kw_before_semi=is_call[],
-                    eq_is_kw_after_semi=is_call[])
+        opts = parse_brackets(ps, K")") do had_commas, had_splat, num_semis, num_subexprs
+            is_call = had_commas || had_splat || initial_semi
+            return (needs_parameters=is_call,
+                    eq_is_kw_before_semi=is_call,
+                    eq_is_kw_after_semi=is_call,
+                    is_call=is_call,
+                    is_block=!is_call && num_semis > 0)
         end
```
NHDaly, line 1179: In my recent allocations reduction PR (https://github.com/JuliaConcurrent/MultiThreadedCaches.jl/pull/10/), I also had to fight a closure that allocates, in order to capture a variable, and forces the variable to be Boxed.

I ended up being able to change the `get!(dict, key) do` pattern into `get(dict, key, :cache_miss)` and then use an if-else instead of a closure in the miss case.
I think it means it'll have to do two different lookups in the dictionary though, which is too bad. :(

Your solution here to return is_call instead of mutating it as a side-effect looks great 👍

c42f: yes in the end I couldn't figure out how to do this with a side effect (even though the `do` block syntax makes that very convenient and natural when looking at the code).  Doing it with a return value was fine in this case because I can modify `parse_brackets` to actually return that return value.

PR #102, opened 2022-09-19
Generate `Expr(:incomplete)`

### LINKED ISSUE #42: When using in REPL, `JuliaSyntax.jl` does not seem to work well with multiline parsing?

c42f:
This allows REPL completion to work correctly. It works by pattern matching the parse tree, rather than hard coding incomplete expression detection into the parser itself.

This took a while because I got distracted with #88. Though doing some of those things first helped make this quite a lot cleaner. There's still more changes from #88 which would help make this nicer but for now it works.

The tests here are somewhat derived from `Base`, but I had to review them all (ie the tests pasted in https://github.com/JuliaLang/JuliaSyntax.jl/pull/71#issuecomment-1225364445) because they turn out to not really make sense.  For example, `"begin;"` and `"begin"` are both the prefix of a block construct, one of them shouldn't come out as `:other`.

Alternative to #71, CC @aviatesk :-)

Fix #42, would close #71

### DIFF
--- a/test/hooks.jl
+++ b/test/hooks.jl
@@ -31,8 +31,69 @@
 
         # Check that Meta.parse throws the JuliaSyntax.ParseError rather than
         # Meta.ParseError when Core integration is enabled.
-        @test_throws JuliaSyntax.ParseError Meta.parse("[x")
+        @test_throws JuliaSyntax.ParseError Meta.parse("[x)")
 
+        JuliaSyntax.enable_in_core!(false)
+    end
+
+    @testset "Expr(:incomplete)" begin
+        JuliaSyntax.enable_in_core!()
+
+        @test Meta.isexpr(Meta.parse("[x"), :incomplete)
+
+        for (str, tag) in [
+                ""             => :none
+                "\""           => :string
+                "\"\$foo"      => :string
+                "#="           => :comment
+                "'"            => :char
+                "'a"           => :char
+                "`"            => :cmd
+                "("            => :other
+                "["            => :other
+                "begin"        => :block
+                "quote"        => :block
+                "let"          => :block
+                "let;"         => :block
+                "for"          => :other
+                "for x=xs"     => :block
+                "function"     => :other
+                "function f()" => :block
+                "macro"        => :other
+                "macro f()"    => :block
+                "f() do"       => :other
+                "f() do x"     => :block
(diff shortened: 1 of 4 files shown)


test/hooks.jl
```diff
@@ -31,8 +31,68 @@
...
+                "mutable struct"    => :other
+                "mutable struct X"  => :block
+                "struct"       => :other
+                "struct X"     => :block
+                "if"           => :other
+                "if x"         => :block
+                "while"        => :other
+                "while x"      => :block
+                "try"          => :block
+                "try x catch"  => :block  # ambiguous - could be :block or :other
```
pfitzseb, line 79: IMHO `:block` is correct here, since `try x catch end` is valid.

c42f: I thought `:other` also fits because `try x catch exc body end` parses. So could either be a block or a symbol for the exception name.  So this seemed ambiguous to me?

c42f: (BTW, I don't think the REPL does much with the distinction here. Though the REPL completion code is a bit hard to follow.)

pfitzseb: `:other` doesn't make sense to me since the presence of `exc` is irrelevant for the validity of the code, but I also don't really know how this value actually ends up being used.

c42f: Seems fair, I'll just change the note next to this to explain the possibility of both syntaxes.  AFAIK this doesn't change things much either way with the REPL (and in any case as noted in the description, I think `Base.incomplete_tag` itself is quite inconsistent in any case).

PR #107, opened 2022-10-02
Rework `JuliaSyntax.parse()` public API

c42f:
`JuliaSyntax.parse()` and `JuliaSyntax.parseall()` were generally pretty inconvenient to use.

This change reworks what I had called `parseall()` to be more similar to `Meta.parse()` and adds `parseall()` and `parseatom()` in analogy to the `Base.Meta` versions of these functions.

The lower level function `parse!()` is provided to work with `ParseStream` for cases where more control is required.

### DIFF
--- a/src/parse_stream.jl
+++ b/src/parse_stream.jl
@@ -160,10 +160,11 @@ const NO_POSITION = ParseStreamPosition(0, 0)
     ParseStream(text::Vector{UInt8},           index::Integer=1; version=VERSION)
     ParseStream(ptr::Ptr{UInt8}, len::Integer, index::Integer=1; version=VERSION)
 
-Construct a `ParseStream` from source `text` which may come in various forms -
-a string, an `IO` object, or a buffer of bytes. In the case that the buffer is
-passed as `ptr,len`, the caller is responsible for preserving the buffer during
-parsing.
+Construct a `ParseStream` from input which may come in various forms:
+* An string (zero copy for `String` and `SubString`)
+* An `IO` object (zero copy for `IOBuffer`). The `IO` object must be seekable.
+* A buffer of bytes (zero copy). The caller is responsible for preserving
+  buffers passed as `(ptr,len)`.
 
 A byte `index` may provided as the position to start parsing.
 
--- a/src/parse_stream.jl
+++ b/src/parse_stream.jl
@@ -166,7 +166,7 @@ Construct a `ParseStream` from input which may come in various forms:
 * A buffer of bytes (zero copy). The caller is responsible for preserving
   buffers passed as `(ptr,len)`.
 
-A byte `index` may provided as the position to start parsing.
+A byte `index` may be provided as the position to start parsing.
 
 ParseStream provides an IO interface for the parser which provides lexing of
 the source text input into tokens, manages insignificant whitespace tokens on
--- a/src/parse_stream.jl
+++ b/src/parse_stream.jl
@@ -155,10 +155,27 @@ const NO_POSITION = ParseStreamPosition(0, 0)
 
 #-------------------------------------------------------------------------------
 """
-ParseStream provides an IO interface for the parser. It
-- Wraps the lexer with a lookahead buffer
-- Removes insignificant whitespace and comment tokens, shifting them into the
-  output implicitly (newlines may be significant depending on `skip_newlines`)
+    ParseStream(text::AbstractString,          index::Integer=1; version=VERSION)
+    ParseStream(text::IO;                                        version=VERSION)
+    ParseStream(text::Vector{UInt8},           index::Integer=1; version=VERSION)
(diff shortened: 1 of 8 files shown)


c42f:
I'm keen for any reviews on this. @pfitzseb if you have time? :-)

src/parser_api.jl
```diff
@@ -111,56 +54,37 @@ function parse(stream::ParseStream; rule::Symbol=:toplevel)
...
 end
 
-function parse(::Type{T}, io::IO;
-               rule::Symbol=:toplevel, version=VERSION, kws...) where {T}
+"""
+    parse!(io::IO; rule=:toplevel, version=VERSION)
+
+Parse Julia source code from a seekable `IO` object. The output is a tuple
+`(tree, diagnostics)`. The `io` position will be set to the next byte of input
+after parsing.
```
pfitzseb, line 62: "to the first unparsed byte of input" or something like that (is that even correct when the whole io content is consumed?)? "next byte" without any context could be kinda confusing.

src/parse_stream.jl
```diff
@@ -155,10 +155,27 @@ const NO_POSITION = ParseStreamPosition(0, 0)
...
+    ParseStream(text::IO;                                        version=VERSION)
+    ParseStream(text::Vector{UInt8},           index::Integer=1; version=VERSION)
+    ParseStream(ptr::Ptr{UInt8}, len::Integer, index::Integer=1; version=VERSION)
+
+Construct a `ParseStream` from source `text` which may come in various forms -
+a string, an `IO` object, or a buffer of bytes. In the case that the buffer is
+passed as `ptr,len`, the caller is responsible for preserving the buffer during
+parsing.
+
+A byte `index` may provided as the position to start parsing.
```
pfitzseb, line 168:
```suggestion
A byte `index` may be provided as the position to start parsing.
```

src/parse_stream.jl
```diff
@@ -155,10 +155,27 @@ const NO_POSITION = ParseStreamPosition(0, 0)
...
-- Wraps the lexer with a lookahead buffer
-- Removes insignificant whitespace and comment tokens, shifting them into the
-  output implicitly (newlines may be significant depending on `skip_newlines`)
+    ParseStream(text::AbstractString,          index::Integer=1; version=VERSION)
+    ParseStream(text::IO;                                        version=VERSION)
+    ParseStream(text::Vector{UInt8},           index::Integer=1; version=VERSION)
+    ParseStream(ptr::Ptr{UInt8}, len::Integer, index::Integer=1; version=VERSION)
+
+Construct a `ParseStream` from source `text` which may come in various forms -
+a string, an `IO` object, or a buffer of bytes. In the case that the buffer is
```
pfitzseb, line 164: I don't think we have an official `IO` interface. Do you think it'd make sense to list the requirements somewhere (in particular, `seek`ability comes to mind)?

src/parser_api.jl
```diff
@@ -65,39 +28,19 @@ Base.display_error(io::IO, err::ParseError, bt) = Base.showerror(io, err, bt)
...
-  `ParseStream` interface can be used to process the output.
-* When `input` is a seekable `IO` subtype, the output is `(tree, diagnostics)`.
-  The buffer `position` will be set to the next byte of input.
-* When `input` is an `AbstractString, Integer`, or `Vector{UInt8}, Integer` the
-  output is `(tree, diagnostics, index)`, where `index` (default 1) is the next
-  byte of input.
+    parse!(stream::ParseStream; rule=:toplevel)
+
+Parse Julia source code from a [`ParseStream`](@ref) object. Output tree data
+structures may be extracted from `stream` with the `build_tree()` function.
```
pfitzseb, line 34:
```suggestion
structures may be extracted from `stream` with the [`build_tree`](@ref) function.
```

pfitzseb (commented):
Looks good to me. Do you think it's worth adding a 1-arg `JuliaSyntax.parse(text) = JuliaSyntax.parse(Expr, text)` convenience functions?

c42f:
Thanks for taking a look!

> Do you think it's worth adding a 1-arg `JuliaSyntax.parse(text) = JuliaSyntax.parse(Expr, text)` convenience functions?

I didn't do this because I'm not sure what the right default is. `Expr` is ok but doesn't offer anything more than `Meta.parse()` so I feel like `SyntaxNode` might be better.

src/parse_stream.jl
c42f: Yes it makes sense. I thought I'd done this, but it's in the docs for `parse!()` I'll put it here too

src/parser_api.jl
c42f: ah this is so hard to word in a way which is clear and correct. How about

"When `parse!` returns, the stream `io` is positioned directly after the last byte which was consumed during parsing."

c42f:
I did a bunch of doc cleanups and a couple of minor tweaks. Should be good to go I guess.

PR #124, opened 2022-10-14
Record fixity of call type in flags

### LINKED ISSUE #99: AST inconsistency between parsing of `=` vs `kw`
There's various cases where `=` is parsed into a `kw` head, but this is inconsistent, especially when named tuples come into play.  This requires various gymnastics in the parser as discussed in https://github.com/JuliaLang/JuliaSyntax.jl#kw-and--inconsistencies

For the named tuple inconsistency, consider the difference between the following:

```julia
julia> dump(Meta.parse("(a=1, b=2)"))
Expr
  head: Symbol tuple
  args: Array{Any}((2,))
    1: Expr
      head: Symbol =
      args: Array{Any}((2,))
        1: Symbol a
        2: Int64 1
    2: Expr
      head: Symbol =
      args: Array{Any}((2,))
        1: Symbol b
        2: Int64 2

julia> dump(Meta.parse("f(a=1, b=2)"))
Expr
  head: Symbol call
  args: Array{Any}((3,))
    1: Symbol f
    2: Expr
      head: Symbol kw
      args: Array{Any}((2,))
        1: Symbol a
        2: Int64 1
    3: Expr
      head: Symbol kw
      args: Array{Any}((2,))
        1: Symbol b
        2: Int64 2
```

Lowering seems to paper over this difference, but it's not very satisfying.

The difference is also confusing when implementing macros as one cannot interpolate expressions like `:(a=1)` into a call and expect them to become keywords. For example:

```julia
julia> :(f($(:(a=1))))
:(f($(Expr(:(=), :a, 1))))

julia> :(f($(Expr(:kw,:x,1))))
:(f(x = 1))
```

Is there a way to resolve this inconsistency? For example, can we remove the `kw` head entirely and just use `=`?  Or can we emit `:kw` in the named tuple case?

One objection to removing `kw` and always using `=` is that interpolating things like `:(a=1)` into a call is that this would change the meaning of the `=` from an assignment into a keyword argument.  However we already have this problem with named tuples.  Alternatively, can we always parse named tuples with `kw` as the expression head?

### LINKED ISSUE #113: Incorrect parsing of `-(a=2)`
On Julia master:
```
julia> @Meta.lower -(a = 2)
:($(Expr(:thunk, CodeInfo(
    @ none within `top-level scope`
1 ─ %1 = Core.get_binding_type(Main, :a)
│   %2 = Base.convert(%1, 2)
│   %3 = Core.typeassert(%2, %1)
│        a = %3
│   %5 = -2
└──      return %5
))))
```
With JuliaSyntax:
```
:($(Expr(:thunk, CodeInfo(
    @ none within `top-level scope`
1 ─ %1 = Core.tuple(:a)
│   %2 = Core.apply_type(Core.NamedTuple, %1)
│   %3 = Core.tuple(2)
│   %4 = (%2)(%3)
│   %5 = Core.kwfunc(-)
│   %6 = (%5)(%4, -)
└──      return %6
))))
```

c42f:
We now record which precise call syntax was used out of the four options:
* Prefix calls with parens
* Prefix operator calls
* Infix operator calls
* Postfix operator calls

This allows us to distinguish keyword arguments from assignment, fixing several bugs with = to kw conversion.

Fixes #113 
Close #99

### DIFF
--- a/src/parse_stream.jl
+++ b/src/parse_stream.jl
@@ -82,10 +82,8 @@ function untokenize(head::SyntaxHead; unique=true, include_flag_suff=true)
         str = str*"-"
         is_trivia(head)  && (str = str*"t")
         is_infix_op_call(head)          && (str = str*"i")
-        # call op flag name mnemonic: i-infix, h,j - left and right of `i` in
-        # alphabetic order
-        is_prefix_op_call(head) && (str = str*"h")
-        is_suffix_op_call(head) && (str = str*"j")
+        is_prefix_op_call(head)  && (str = str*"pre")
+        is_postfix_op_call(head) && (str = str*"post")
         has_flags(head, TRIPLE_STRING_FLAG) && (str = str*"s")
         has_flags(head, RAW_STRING_FLAG) && (str = str*"r")
         is_suffixed(head) && (str = str*"S")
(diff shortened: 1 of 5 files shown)


src/parse_stream.jl
```diff
@@ -70,7 +81,11 @@ function untokenize(head::SyntaxHead; unique=true, include_flag_suff=true)
     if include_flag_suff && suffix_flags != EMPTY_FLAGS
         str = str*"-"
         is_trivia(head)  && (str = str*"t")
-        is_infix(head)   && (str = str*"i")
+        is_infix_op_call(head)          && (str = str*"i")
+        # call op flag name mnemonic: i-infix, h,j - left and right of `i` in
+        # alphabetic order
+        is_prefix_op_call(head) && (str = str*"h")
+        is_suffix_op_call(head) && (str = str*"j")
```
pfitzseb, line 88: We're not limited to one-char suffixes here, right? So `-pre`/`-suf` or `-pre`/`-post` would probably be the more obvious choices.

c42f: Flags can be combined (eg raw and string), so I did limit it to one char suffixes on purpose.

Could be a good idea though to make these clearer. But if we do, let's do all the flags at once.

c42f: (ie, in a separate PR)

c42f: Actually I'll do this now after all: These particular flags aren't combined with others so this will be clearer and the lack of consistency shouldn't be a big deal.

c42f: Done

PR #127, opened 2022-10-21
Wrap `var""` nonstandard identifiers in var nodes

c42f:
Useful to hold associated trivia (delimiter and var prefix) and indicate in a clean way that `var""` was used.

Part of #88

### DIFF
--- a/src/expr.jl
+++ b/src/expr.jl
@@ -30,7 +30,14 @@ function _to_expr(node::SyntaxNode; iteration_spec=false, need_linenodes=true,
         end
     end
     nodekind = kind(node)
-    if nodekind == K"?"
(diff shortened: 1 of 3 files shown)


c42f:
Hmm ok. So `k == K"Identifier"` doesn't work for identifiying identifiers after this change.

c42f:
So... I feel like
* A `K"var"` head is a nice unambiguous representation of the concrete syntax
* A plain symbol (as exists currently and as `Expr` uses) is clearly a nice representation of the abstract syntax
* An internal node with head `K"Identifier"` would be a weird halfway solution between these

So let's go with `K"var"` head I guess

PR #131, opened 2022-10-28
Permit parens in function call signatures

### LINKED ISSUE #116: Incorrect parsing of `function` with parens.
master:
```
julia> dump(:(function (f(::T) where {T}) end))
Expr
  head: Symbol function
  args: Array{Any}((2,))
    1: Expr
      head: Symbol where
      args: Array{Any}((2,))
        1: Expr
          head: Symbol call
          args: Array{Any}((2,))
            1: Symbol f
            2: Expr
              head: Symbol ::
              args: Array{Any}((1,))
                1: Symbol T
        2: Symbol T
    2: Expr
      head: Symbol block
      args: Array{Any}((2,))
        1: LineNumberNode
          line: Int64 1
          file: Symbol REPL[11]
        2: LineNumberNode
          line: Int64 1
          file: Symbol REPL[11]
```
JuliaSyntax
```
julia> dump(:(function (f(::T) where {T}) end))
Expr
  head: Symbol function
  args: Array{Any}((2,))
    1: Expr
      head: Symbol tuple
      args: Array{Any}((1,))
        1: Expr
          head: Symbol where
          args: Array{Any}((2,))
            1: Expr
              head: Symbol call
              args: Array{Any}((2,))
                1: Symbol f
                2: Expr
            2: Symbol T
    2: Expr
      head: Symbol block
      args: Array{Any}((2,))
        1: LineNumberNode
          line: Int64 1
          file: Symbol REPL[21]
        2: LineNumberNode
          line: Int64 1
          file: Symbol REPL[21]
```

c42f:
This permits the extra parentheses in things like

    function (funcname(some, long, argument, list)
              where {Type,Params})
        body
    end

This syntax "works" in the reference parser and has been seen in the wild so we need to support it for compatibility. (However, it's broken when combined with return type annotations which suggests it's more a syntactic aberration rather than an intentional feature.)

Fix #116

### DIFF
--- a/src/kinds.jl
+++ b/src/kinds.jl
@@ -880,6 +880,7 @@ const _kind_names =
         "tuple"
         "ref"
         "vect"
+        "parens"
         # Concatenation syntax
         "braces"
         "bracescat"
--- a/src/parse_stream.jl
+++ b/src/parse_stream.jl
@@ -528,6 +528,46 @@ function peek_behind(stream::ParseStream, pos::ParseStreamPosition)
     end
 end
 
+function first_child_position(stream::ParseStream, pos::ParseStreamPosition)
+    # Find the first nontrivia range which is a child of this range but not a
+    # child of the child
+    c = 0
+    @assert pos.range_index > 0
+    parent = stream.ranges[pos.range_index]
+    i = pos.range_index-1
+    while i >= 1
+        if stream.ranges[i].first_token >= parent.first_token &&
+               (c == 0 || stream.ranges[i].first_token < stream.ranges[c].first_token) &&
+               !is_trivia(stream.ranges[i])
+            c = i
+        end
+        i -= 1
+    end
+
+    # Find first nontrivia token
+    t = 0
+    for i = parent.first_token:parent.last_token
+        if !is_trivia(stream.tokens[i])
+            t = i
+            break
+        end
+    end
+
+    if c != 0
+        if t != 0
+            if stream.ranges[c].first_token > t
+                return ParseStreamPosition(t, c-1)
+            else
+                return ParseStreamPosition(stream.ranges[c].last_token, c)
+            end
+        else
+            return ParseStreamPosition(stream.ranges[c].last_token, c)
+        end
+    else
+        return ParseStreamPosition(t, c)
+    end
+end
+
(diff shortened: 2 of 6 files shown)


src/kinds.jl
```diff
@@ -880,6 +880,7 @@ const _kind_names =
         "tuple"
         "ref"
         "vect"
+        "parens"
```
c42f, line 883: This is here because I wanted to use it in the test harness, and because I expect to use it in the main parser later on, as part of #88

pfitzseb:
> However, it's broken when combined with return type annotations which suggests it's more a syntactic aberration rather than an intentional feature.

Do you mean
```
function (funcname(some, long, argument, list)
          where {Type,Params})::Int
    body
end
```
If so, I don't think it's reasonable to expect that to work. The usual way of specifying return types via
```
function ((funcname(some, long, argument, list)::Int)
          where {Type,Params})::Int
    body
end
```
works fine afterall, even with the additional parens.

c42f:
I mean you can't "just add grouping parens" without changing the relative precedence of `::` and `where`:

Ie, we can't go from the normal syntax

```julia
julia> function f(a::T)::T where T
           a
       end
f (generic function with 1 method)

julia> f(1)
1
```

To the same thing with parens without the precedence of `::` and `where` changing:

```julia
julia> function (g(a::T)::T where T)
           a
       end
ERROR: UndefVarError: T not defined
```

Additional parens (like in your example) are required to manually fix the precedence:

```julia
julia> function ((h(a::T)::T) where T)
           a
       end
h (generic function with 1 method)
```

all of which indicates to me that allowing parens here at all is probably just a bad idea. (Must be allowed for compatibility, though we could warn in future?)

c42f:
I'm going to merge this now - the comment about this being bad syntax is just my opinion that this PR doesn't take a stance on (other than complaining about it in a code comment hehe)

oscardssmith:
So fun fact: This broke parsing of
```
function (::A(B))()
1
end
 ```

oscardssmith:
(This matters for ChainRulesCore https://github.com/JuliaDiff/ChainRulesCore.jl/blob/9c8fcd22ee1432bb1b1707215e09af4a1ae3b072/src/projection.jl#L34 which means that with this change pretty much every library that defines chain rules is broken with this).

PR #133, opened 2022-10-31
More flattened form for multiple frakentuple parameters

c42f:
Here we emit `(a, b; c, d; e, f)` as

    (tuple a b (parameters c d) (parameters e f))

which should make these simpler to process, rather than using the nested form

    (tuple a b (parameters c d parameters e f)))

Part of #88


c42f:
CC @BenChung we discussed this previously, I think.

PR #146, opened 2022-11-04
Lower `@ .` to `@ __dot__` not in parser but in `Expr` conversion

c42f:
The reason to use `__dot__` is to allow `macro __dot__` to be defined in normal Julia source. But doing this in the parser is an awkward special case - better to do it later during some lowering step.  Here done in `Expr` conversion.


c42f:
Eh let's merge this so I can fix some other things on top of it. Presumably it's non-controversial :)

PR #151, opened 2022-11-20
Parse dotted calls with `dotcall` head

### LINKED ISSUE #38: `.<:` and `.>:` should not be parsed as unary operators
From https://github.com/JuliaLang/JuliaSyntax.jl/issues/31#issuecomment-1164129041_

* Unary versions `.<:` and `.>:` parse but they shouldn't.
* Also, them not being translated to the dotted versions in the `Expr` conversion is weird.

```
julia> JuliaSyntax.parseall(Expr, ".<: b", rule=:statement)
:(<:b)

julia> Meta.parse(".<: b")
ERROR: Base.Meta.ParseError("\".<:\" is not a unary operator")
Stacktrace:
...

julia> JuliaSyntax.parseall(Expr, ".>: b", rule=:statement)
:(>:b)

julia> Meta.parse(".>: b")
ERROR: Base.Meta.ParseError("\".>:\" is not a unary operator")
...
```

### LINKED ISSUE #90: Rework handling of `.` tokenization and operator calls
The handling of `.` in the tokenizer / parser is pretty wonky / inconsistent because the tokenization of `.` is context-dependent.

`bump_split()`, in particular is quite ugly and shouldn't exist:
* `...` is tokenized into `K"..."` which is usually correct.  But incorrect for `import`/`using` statements as in `import ...A   ==> (import (. . . . A))`, necessitating the splitting of tokens with `bump_split` - this is ugly!
* Dotted operators are tokenized with the dot as part of the operator. But sometimes we have to split them, as in standalone dotted operators `.+    ==>  (. +)`

In addition, `Expr` is quite inconsistent about dotted infix calls vs dotted prefix calls. We should really fix this?  Then we can remove the `.` from the operator names and treat it as separate syntax as it should be!  (See also #88)

```julia
julia> dump(Meta.parse("a .+ b"))
Expr
  head: Symbol call
  args: Array{Any}((3,))
    1: Symbol .+
    2: Symbol a
    3: Symbol b

julia> dump(Meta.parse("f.(a, b)"))
Expr
  head: Symbol .
  args: Array{Any}((2,))
    1: Symbol f
    2: Expr
      head: Symbol tuple
      args: Array{Any}((2,))
        1: Symbol a
        2: Symbol b
```

c42f:
Dotted call syntax parses into various forms which aren't really consistent. Especially, Expr is inconsistent about dotted infix calls vs dotted prefix calls. In this change we adopt a more consistent (and hopefully less mysterious!) parsing where dotted calls get their own `dotcall` head which is otherwise like the `call` head:

    f.(a, b)  ==>  (dotcall f a b)
    a .+ b    ==>  (dotcall-i a + b)
    .+ b      ==>  (dotcall-pre + b)

In other cases where a dotted operator appears as an atom we split the dot from an operator, so `.+` becomes `(. +)`. Thus, for comparison chains and prefix calls, we have the following parsings:

    a .< b < c  ==>  (comparison a (. <) b < c)
    .*(a,b)     ==>  (call (. *) a b)

TODO: Unambiguously reconstructing the associated Expr in the latter cases is more challenging than expected and still requires some work.

Fix #90
Also incidentally fix #38

### DIFF
--- a/src/expr.jl
+++ b/src/expr.jl
@@ -31,7 +31,7 @@ function reorder_parameters!(args, params_pos)
 end
 
 function _to_expr(node::SyntaxNode; iteration_spec=false, need_linenodes=true,
-                  eq_to_kw=false, inside_dot_expr=false, inside_vect_or_braces=false)
+                  eq_to_kw=false, inside_vect_or_braces=false)
     if !haschildren(node)
         val = node.val
         if val isa Union{Int128,UInt128,BigInt}
@@ -125,11 +125,9 @@ function _to_expr(node::SyntaxNode; iteration_spec=false, need_linenodes=true,
         args[2] = _to_expr(node_args[2])
     else
         eq_to_kw_in_call =
-            headsym == :call && is_prefix_call(node)          ||
+            ((headsym == :call || headsym == :dotcall) && is_prefix_call(node)) ||
             headsym == :ref
-        eq_to_kw_all = headsym == :parameters && !inside_vect_or_braces ||
-                      (headsym == :tuple && inside_dot_expr)
-        in_dot = headsym == :.
+        eq_to_kw_all = headsym == :parameters && !inside_vect_or_braces
         in_vb = headsym == :vect || headsym == :braces
         if insert_linenums && isempty(node_args)
             push!(args, source_location(LineNumberNode, node.source, node.position))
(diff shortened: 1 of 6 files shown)


c42f:
To quote the latest rebased commit message, I've punted on some non-dotcall related dotted operator splitting for now:

"There's other cases where it would also be consistent to split the dot
from the operator, but these are more challenging to convert back to a
compatible Expr so I've punted on these for now. For example, we'd like
    
    .*(a,b)     ==>  (call (. *) a b)
    .+(a,)      ==>  (call (. +) a)
    
but these are not yet implemented as we need to be able to distinguish
them from the likes of `(.+)(a,)` which the reference parser treats
differently from `.+(a,)`"

PR #182, opened 2023-01-15
Support for running JuliaSyntax on Julia 1.0

### LINKED ISSUE #180: Runtime Julia 1.0 support
The package currently has Julia 1.6 as the min version. Is that essential? Or could it be made to work on Julia 1.0 as well?

I've been looking into adding this to the language server so that we can surface the fancy cool error messages, but that would only work if it runs on Julia 1.0 as well...

Eventually I hope to statically compile the language server, and at that point runtime support for old Julia versions would not be required anymore, but my sense is that that is not around the corner ;)

c42f:
Mostly just compat changes here, adding a bunch of Base utility functions we use which don't exist in earlier Julia versions.

There's still a few failures in the extended tests, probably due to differences in line number node placement and the like in the reference parser.

Fix #180

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,7 +13,14 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - '1.0'
+          - '1.1'
+          - '1.2'
+          - '1.3'
+          - '1.4'
+          - '1.5'
           - '1.6'
+          - '1.7'
           - '1'
           - 'nightly'
         os:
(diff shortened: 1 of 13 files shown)


davidanthoff:
@c42f do you think it might make sense to just disable those tests for older Julia versions?

c42f:
Yea. I was working on doing that in a surgical way, but I'm really not sure it's worth the effort at this point, to be honest.

Older Julia versions have some weird placing of line number nodes and some bugs in the reference parser (eg broken triple quoted string deindentation) which make comparison with the reference parser from those old versions quite annoying.

davidanthoff:
Hurray! Could you also tag a new release?

My plan is to start incorporating this into the language server, simply for the nice syntax error messages. Does that sound good to you?

c42f:
Sounds great to me!

New release tagged: https://github.com/JuliaRegistries/General/pull/76003

PR #183, opened 2023-01-18
Make things compabitle with Julia 1.0

### LINKED ISSUE #180: Runtime Julia 1.0 support

davidanthoff:
Fixes #180.

Right now I'm stuck with a bug in hooks.jl, namely `Core.Compiler.fl_parse` does not exist in Julia 1.0. @c42f, any idea how we can get around that? The goal here is to make simply make it possible to run JuliaSyntax in the LanguageServer, so no need for any of the stuff that replaces things in Julia itself.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,6 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - '1.0'
           - '1.6'
           - '1'
           - 'nightly'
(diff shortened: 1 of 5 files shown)


davidanthoff:
Dang, didn't see https://github.com/JuliaLang/JuliaSyntax.jl/pull/182 :) Closing this one here :)

c42f:
Haha oops!  You were well on the way to fixing the core problems in this PR, thanks for diving in anyway :)

PR #191, opened 2023-02-11
Add lineno to SourceFile

### LINKED ISSUE #190: Add `lineno` to `SourceFile`?
https://github.com/JuliaDebug/Cthulhu.jl/pull/345 uses JuliaSyntax to represent source code with expression boundaries mapped to character positions within the source text. In that application, [CodeTracking](https://github.com/timholy/CodeTracking.jl) is used to extract source-text for specific methods which may be partway into a file.

Question: for "snippets" like these, should the starting line number `lineno` be added as a new field to `SourceFile`? Or do you need it to represent "a whole file"? If so, then adding it as a field seems incorrect, since of course a whole file will start at line 1.

timholy:
Closes #190

### DIFF
--- a/src/source_files.jl
+++ b/src/source_files.jl
@@ -1,7 +1,9 @@
 """
-    SourceFile(code [; filename=nothing])
+    SourceFile(code [; filename=nothing, lineno=1])
 
-A UTF-8 source code string with associated file name and indexing structures.
+A UTF-8 source code string with associated file name and line number.
+
+`SourceFile` stores the character positions of line starts to facilitate indexing.
 """
 struct SourceFile
     # We use `code::String` for now but it could be some other UTF-8 based
@@ -11,11 +13,12 @@ struct SourceFile
     # https://en.wikipedia.org/wiki/Rope_(data_structure)
     code::String
     filename::Union{Nothing,String}
+    lineno::Int
     # String index of start of every line
     line_starts::Vector{Int}
 end
 
-function SourceFile(code::AbstractString; filename=nothing)
+function SourceFile(code::AbstractString; filename=nothing, lineno=1)
     line_starts = Int[1]
     for i in eachindex(code)
         # The line is considered to start after the `\n`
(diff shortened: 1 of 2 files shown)


src/source_files.jl
```diff
@@ -1,7 +1,9 @@
 """
-    SourceFile(code [; filename=nothing])
+    SourceFile(code [; filename=nothing, lineno=1])
```
c42f, line 2: Can we call `lineno` something more explanatory. Maybe `first_line`?

src/source_files.jl
```diff
@@ -25,31 +28,32 @@ function SourceFile(code::AbstractString; filename=nothing)
...
 
 # Get line number of the given byte within the code
-function source_line(source::SourceFile, byte_index)
-    line = searchsortedlast(source.line_starts, byte_index)
-    return (line < lastindex(source.line_starts)) ? line : line-1
+function source_line_index(source::SourceFile, byte_index)
+    lineidx = searchsortedlast(source.line_starts, byte_index)
+    return (lineidx < lastindex(source.line_starts)) ? lineidx : lineidx-1
 end
+source_line(source::SourceFile, byte_index, lineidx=source_line_index(source, byte_index)) = lineidx + source.lineno - 1
```
c42f, line 43: We don't want the extra parameter to be part of the public API do we?

```suggestion
_source_line(source::SourceFile, lineidx) = lineidx + source.lineno - 1
source_line(source::SourceFile, byte_index) = _source_line(source, source_line_index(source, byte_index))
```

src/source_files.jl
```diff
@@ -25,31 +28,32 @@ function SourceFile(code::AbstractString; filename=nothing)
...
-    i = source.line_starts[line]
+    lineidx = source_line_index(source, byte_index)
+    i = source.line_starts[lineidx]
     column = 1
     while i < byte_index
         i = nextind(source.code, i)
         column += 1
     end
-    line, column
+    source_line(source, byte_index, lineidx), column
```
c42f, line 56: If we include the suggestion above we need this too

```suggestion
    _source_line(source, lineidx), column
```

c42f (commented):
I think this is good :+1:

I'd prefer a more specific name than `lineno`. Perhaps `first_line`?

c42f (approved):
Looks good to me.

c42f:
@timholy I added you to this repo so feel free to merge etc

c42f:
I took the liberty of editing your description so that #191 doesn't get auto-closed

PR #193, opened 2023-02-11
Split SyntaxNode into TreeNode & SyntaxData

### LINKED ISSUE #192: Location of TypedSyntaxNode
I now think that the best way to handle https://github.com/JuliaDebug/Cthulhu.jl/pull/345 is to create
```julia
mutable struct TypedSyntaxNode <: AbstractSyntaxNode
    node::SyntaxNode
    typ::Any
    # These duplicate info in SyntaxNode but are needed so we can build trees of TypedSyntaxNodes
    parent::Union{Nothing,SyntaxNode}
    val::Any
end
```

Do you want that type, possibly together with its printing functions, to live here? Or should it be a separate package?

A more invasive but extensible approach would be to factor out the "data" of nodes from the tree structure. Then assembly and traversal become orthogonal from content. So we'd have `mutable struct TreeNode{NodeData}` with `parent`, `children`, and `val::Union{NodeData,Nothing}`, where `val` (for any node that isn't just a container of other nodes) might be
```julia
struct SyntaxData
    source::SourceFile
    raw::GreenNode{SyntaxHead}
    position::Int
end
```

Then we'd only need `TypedSyntaxData` with an extra `typ` field; everything else would be reusable.

timholy:
Closes #192

Pleasantly straightforward. Note that this object is slightly bigger since `val` has basically become two fields now, one for children and one for the actual value. I hope that's a tolerable increase.

I haven't timed anything to see if there are any regressions. (If constprop works, I think there won't be any.) Is there a task I should use?

### DIFF
--- a/src/syntax_tree.jl
+++ b/src/syntax_tree.jl
@@ -9,6 +9,24 @@ mutable struct TreeNode{NodeData}   # ? prevent others from using this with Node
     data::Union{Nothing,NodeData}
 end
 
+# Implement "pass-through" semantics for field access: access fields of `data`
+# as if they were part of `TreeNode`
+function Base.getproperty(node::TreeNode, name::Symbol)
+    name === :parent && return getfield(node, :parent)
+    name === :children && return getfield(node, :children)
+    d = getfield(node, :data)
+    name === :data && return d
+    return getproperty(d, name)
+end
+
+function Base.setproperty!(node::TreeNode, name::Symbol, x)
+    name === :parent && return setfield!(node, :parent, x)
+    name === :children && return setfield!(node, :children, x)
+    name === :data && return setfield!(node, :data, x)
+    d = getfield(node, :data)
+    return setfield!(d, name, x)
+end
+
 const AbstractSyntaxNode = TreeNode{<:AbstractSyntaxData}
 
 struct SyntaxData <: AbstractSyntaxData
@@ -128,14 +146,6 @@ function SyntaxNode(source::SourceFile, raw::GreenNode{SyntaxHead}, position::In
         end
         return node
     end
-end
-
-function Base.getproperty(node::TreeNode, name::Symbol)
-    name === :parent && return getfield(node, :parent)
-    name === :children && return getfield(node, :children)
-    d = getfield(node, :data)
-    name === :data && return d
-    return getproperty(d, name)
 end
 
 haschildren(node::TreeNode) = node.children !== nothing
--- a/src/syntax_tree.jl
+++ b/src/syntax_tree.jl
@@ -16,12 +16,6 @@ end
 
 const SyntaxNode = TreeNode{SyntaxData}
 
-"""
-Design options:
-* rust-analyzer treats their version of an untyped syntax node as a cursor into
-  the green tree. They deallocate aggressively.
-"""
-
 # Value of an error node with no children
 struct ErrorVal
 end
--- a/src/syntax_tree.jl
+++ b/src/syntax_tree.jl
@@ -1,19 +1,26 @@
 #-------------------------------------------------------------------------------
 # AST interface, built on top of raw tree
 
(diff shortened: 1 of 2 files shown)


src/syntax_tree.jl
```diff
@@ -1,20 +1,27 @@
...
 end
 
+const SyntaxNode = TreeNode{SyntaxData}
+
+"""
+Design options:
+* rust-analyzer treats their version of an untyped syntax node as a cursor into
+  the green tree. They deallocate aggressively.
+"""
+
```
c42f, line 24:
```suggestion
```

This is a pretty old comment, let's just delete it.

c42f (commented):
This seems pretty reasonable. But I'm curious how much this helps you downstream?  It looks like you don't get much functionality with `TreeNode{SomeOtherNodeData}` as the methods are still specialized to `SyntaxNode` here.

In terms of benchmarks, there's not really any formal benchmarks at this point.  As a rough guide, converting all the way to `Expr` should be about 5x faster than the existing parser. You could try the following code before and after the changes? (I'll PR this to test/benchmark.jl shortly. I'm not sure what reliable infrastructure we have for benchmarking automatically...)

```julia
using BenchmarkTools
using JuliaSyntax

include("test_utils.jl")

function concat_base()
    basedir = joinpath(Sys.BINDIR, "..", "share", "julia", "base")
    io = IOBuffer()
    for f in find_source_in_path(basedir)
        write(io, read(f, String))
        println(io)
    end
    return String(take!(io))
end

all_base_code = concat_base()

b_ParseStream = @benchmark JuliaSyntax.parse!(JuliaSyntax.ParseStream(all_base_code), rule=:toplevel)
b_GreenNode   = @benchmark JuliaSyntax.parseall(JuliaSyntax.GreenNode, all_base_code)
b_SyntaxNode  = @benchmark JuliaSyntax.parseall(JuliaSyntax.SyntaxNode, all_base_code)
b_Expr        = @benchmark JuliaSyntax.parseall(Expr, all_base_code)

@info "Benchmarks" ParseStream=b_ParseStream GreenNode=b_GreenNode SyntaxNode=b_SyntaxNode Expr=b_Expr
```

timholy:
> But I'm curious how much this helps you downstream?

We might still need to broaden some other methods but I was planning to do that focally as I build out functionality. For now this is all I need, and it seems to clearly pave the way for elegantly handling both typed and untyped syntax trees.

timholy:
Thanks for the benchmarks; the last three are unchanged, but surprisingly there's a ~30% regression on the first one, which drops to ~15% if I add `Base.@constprop :aggressive` in front of `getproperty`. Personally I wouldn't worry about that, but you might, so I thought I'd better raise it.

c42f:
> there's a ~30% regression on the first one

That would be troubling, but it's very confusing because `SyntaxNode` shouldn't be involved in that benchmark!

c42f (approved):
I think you should go ahead and merge this if you think it's the right way forward.

It's surprisingly non-breaking :heart:

c42f (approved):

c42f:
(I would like to understand whether the performance regression is real, it does bother me and it should be impossible. But I don't know if I have space in my head to dig deep into it right now.)

timholy:
Before I merge:
- [ ] understand the performance regression
- [ ] check test coverage and see if it needs to be improved before it's safe to merge this refactor (the most recent commit has some `val` -> `children` changes that arguably should have caused test failures in previous iterations of this PR).

timholy:
I'm not quite sure what was happening before, but today when I run the benchmarks there are no signs of trouble.

c42f:
> today when I run the benchmarks there are no signs of trouble

Cool this makes sense! The only thing I could think of was that perhaps the `getproperty` overloading was causing some weird inference limit to be hit somewhere in supposedly unrelated code. But that seemed like a long shot.

Issue #134, opened 2022-11-01
Failures parsing General registry


c42f:
Here's a list of failures when parsing General restricted to cases where
* JuliaSyntax.jl fails to parse a file
* the reference parser accepts the file

### Parser failures
* [ ] `A[x ? y : end]` - AssociativeArrays_0.0.3/src/namedaxis.jl, Reduce_1.2.12/src/Reduce.jl, UnROOT_0.8.1/src/UnROOT.jl
* [ ] `function (f()) where T end` - DMRJtensor_0.8.7/lib/Qtensor.jl, DMRJtensor_0.8.7/lib/decompositions.jl, EnglishText_0.6.0/src/numeric.jl, ExprTools_0.1.8/test/function.jl, MemoryMutate_0.0.4/src/MemoryMutate.jl, MultivariatePolynomials_0.4.2/src/differentiation.jl, Singular_0.9.1/src/module/vector.jl, etc why people why??
* [ ] `.1..0.1`  -  DomainSets_0.5.9/test/test_setoperations.jl, `0x01..0xcf` - Rasters_0.2.0/test/sources/gdal.jl
* [ ] `import Base.:*` - Empirikos_0.4.5/src/ebayes_targets.jl, `import A.:⋅` - MeasureTheory_0.13.2/test/runtests.jl, Pathnames_0.1.0/src/Pathnames.jl, Porta_0.1.4/src/geometry/biquaternions.jl, `Base.(:*)` - UnitfulLatexify_1.5.2/src/UnitfulLatexify.jl
* [ ] `function ($M).$J() end` - Knet_1.4.9/src/cuarrays/reduction.jl
* [ ] `@a@b` - MathematicalSystems_0.11.11/test/@system.jl
* [ ] `[a~b]`- ModelingToolkitStandardLibrary_0.1.0/test/test_continuous.jl, NeuralPDE_4.4.0/test/NNPDE_tests.jl, etc
* [ ] `@(A)` - NumericalMethodsforEngineers_2.0.0/examples/ch07/WRM/Fig.7.14.jl
* [ ] `-1.#comment` - POMDPModels_0.4.15/src/MountainCar.jl
* [ ] `A.@a(x).b` - WorkerUtilities_1.0.1/test/runtests.jl

### Expr conversion failures
* [ ] JuliaFormatter_0.22.0/test/files/PProf.jl
* [ ] MathOptInterfaceMosek_0.5.2/examples/chainsing.jl
[...]

PR #197, opened 2023-02-15
Fix for operator-named macros: parse `@+x` as `@+ x`

c42f:
Allow operator-named macros to be used without spaces.

Part of #134


c42f:
Welp, it looks simple, but gosh did I spend a long time trying to figure out how to fix this one neatly.  Passes tests... so let's go.

PR #199, opened 2023-02-17
Prototype: Underscores and prefix anonymous function syntax

c42f:
As discussed in https://github.com/JuliaLang/julia/issues/38713

### DIFF
--- a/src/expr.jl
+++ b/src/expr.jl
@@ -30,6 +30,27 @@ function reorder_parameters!(args, params_pos)
     insert!(args, params_pos, pop!(args))
 end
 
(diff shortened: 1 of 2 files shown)


adienes:
is #199 complementary, incompatible, or orthogonal to this proposal? I must admit I am not at all a fan of the aesthetics of piping into a headless ` |> ->(...)` compared to `/>` or `/>>` (although I know they have different semantics)

c42f:
I assume you mean #148?

It's complimentary and I've merged the changes from this PR into there as well, to see how they'd play out together.

c42f:
I doubt we're ever doing this...

PR #201, opened 2023-02-17
Support `begin/end` indexing of `SourceFile`

timholy:


test/source_files.jl
```diff
@@ -26,4 +26,14 @@
         @test source_location(SourceFile(; filename=path), 1) == (1,1)
         @test source_location(SourceFile(; filename=path, first_line=7), 1) == (7,1)
     end
+
+    @test SourceFile("a\nb\n")[1:2] == "a\n"
+    @test SourceFile("a\nb\n")[3:end] == "b\n"
+    if Base.VERSION >= v"1.4"
+        # Protect the `[begin` from being viewed by the parse on older Julia versions
+        @test eval(Meta.parse("""SourceFile("a\nb\n")[begin:end]""")) == "a\nb\n"
```
c42f, line 34: hehehe we *are* the parser!

So we can almost use `JuliaSyntax.parse(Expr, """SourceFile("a\nb\n")[begin:end]""", version=v"1.6")` unconditionally here.

But alas, lowering won't understand the result with the `begin` embedded in the `Expr`... so not really :sweat_smile:

c42f (approved):
looks good to me

test/source_files.jl
timholy: It is indeed quite ironic! I guess if this were to swap itself in as the `include` parser, then even older Julia versions would be prepared for this. But given the necessary changes in https://github.com/JuliaLang/julia/pull/46372, presumably this won't happen.

c42f: `JuliaSyntax.enable_in_core!()` does make it work with `include()`!

But only in Julia 1.6+ because it requires the hooks implemented in https://github.com/JuliaLang/julia/pull/35243

PR #204, opened 2023-02-18
Improve test coverge of source_files, syntax_tree

timholy:
These two files are particularly affected by recent and upcoming
changes (e.g., #193). This adds a bit more coverage as a guard
against breakage.

### DIFF
--- a/src/syntax_tree.jl
+++ b/src/syntax_tree.jl
@@ -286,10 +286,10 @@ end
 """
 Print the code, highlighting the part covered by `node` at tree `path`.
 """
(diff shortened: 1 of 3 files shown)


timholy:
Overall coverage is quite good, but on `main` these two files are on the lower side (though still not too bad): https://app.codecov.io/gh/JuliaLang/JuliaSyntax.jl/tree/main/src

c42f (approved):
Awesome, thanks!

Yes the coverage is a bit ... ahem ... uneven.

Parser coverage should be good, but coverage for the errors and diagnostics machinery needs work.

PR #208, opened 2023-02-23
AbstractSyntaxNode generalization & convenience

timholy:
These add a couple more convenience methods useful for TypedSyntax.


timholy:
I *think* these are the last obvious additions needed by TypedSyntax. What are your thoughts on a new release?

c42f:
If you need it released, let's release 0.3.1 right away!

PR #209, opened 2023-02-23
Bump to 0.3.1 and remove old Printf dependency

c42f:
Based on a quick look through the changes I think these are all fixes and generalizations so we should be compatible with 0.3


timholy:
It's a testament to how you designed this that I was able to add what I needed without needing a breaking release!

c42f:
Aww thanks Tim!

I don't know if that's true, but I'm glad (and kinda surprised) that it worked out so neatly. Thanks for the contributions :-)

PR #210, opened 2023-02-24
Support `copy` for TreeNodes

timholy:


c42f (approved):
Initially I thought the combination of shallow and deep copy was weird. But it's a reasonable analogy to how array copy copies the container but not the data. And it's pretty much how `copy(::Expr)` works too.  So LGTM :)

src/syntax_tree.jl
```diff
@@ -244,6 +244,21 @@ function Base.push!(node::SN, child::SN) where SN<:AbstractSyntaxNode
     push!(args, child)
 end
 
+function Base.copy(node::TreeNode)
+    # copy is deep in the tree, shallow in the data
```
c42f, line 248:
```suggestion
    # copy the container but not the data (ie, deep copy the tree, shallow copy the data). copy(::Expr) is similar
```

timholy:
OK if I tag a new release after this merges? (Or feel free to do so yourself.)

c42f:
Go for it :+1:

PR #212, opened 2023-03-03
Experiment: Make macros bind stronger than commas within parentheses

c42f:
As discussed in https://github.com/JuliaLang/julia/issues/36547

That is, `f(@x a, b)` parses as

    (call f (macrocall a) b)

rather than

    (call f (macrocall (tupel a b)))

Unfortunately, this is mildly breaking, due to such strange syntax as `(@unpack a,b = c)`

### DIFF
--- a/src/parse_stream.jl
+++ b/src/parse_stream.jl
@@ -236,12 +236,16 @@ mutable struct ParseStream
     diagnostics::Vector{Diagnostic}
     # Counter for number of peek()s we've done without making progress via a bump()
     peek_count::Int
+
+    # Feature flags
     # (major,minor) version of Julia we're parsing this code for.
     # May be different from VERSION!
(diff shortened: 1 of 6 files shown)


nsajko:
Perhaps it'd make sense to deprecate `f(@x a, b)` and/or introduce warnings, but changing the semantics from one valid interpretation to another would be horrible.

c42f:
> changing the semantics from one valid interpretation to another would be horrible

It's "horrible", but evidence suggests that users don't understand the existing precedence and actively write bugs with the existing semantics. Look at [the analysis here](https://github.com/JuliaLang/julia/issues/36547)

If we fix more cases in the ecosystem than we break by making this change, is that actually "horrible"?  There's no right answer, only tradeoffs.

PR #213, opened 2023-03-08
add a very basic precompile workload

KristofferC:
I want to use this package in OhMyREPL but the current latency is too big. This PR makes it as good as Tokenize.jl

Before:

```
julia> @time @eval collect(JuliaSyntax.Tokenize.Lexer("1+1 == 2"))
  1.083684 seconds (1.04 M allocations: 63.173 MiB, 0.86% gc time, 99.98% compilation time)
```

After:

```
julia> @time @eval collect(JuliaSyntax.Tokenize.Lexer("1+1 == 2"))
  0.016808 seconds (29.64 k allocations: 1.939 MiB, 62.47% compilation
```

cc @timholy

### DIFF
--- a/src/precompile.jl
+++ b/src/precompile.jl
@@ -1,6 +1,5 @@
 # Just parse some file as a precompile workload
 let filename = joinpath(@__DIR__, "literal_parsing.jl")
     text = read(filename, String)
-    stream = JuliaSyntax.ParseStream(text)
-    JuliaSyntax.parse!(stream)
+    parseall(Expr, text)
 end
--- a/src/precompile.jl
+++ b/src/precompile.jl
@@ -1,0 +1,5 @@
+# Just parse some file as a precompile workload
+filename = joinpath(@__DIR__, "literal_parsing.jl")
+text = read(filename, String)
+stream = JuliaSyntax.ParseStream(text)
+JuliaSyntax.parse!(stream)
--- a/src/precompile.jl
+++ b/src/precompile.jl
@@ -1,5 +1,6 @@
 # Just parse some file as a precompile workload
-filename = joinpath(@__DIR__, "literal_parsing.jl")
-text = read(filename, String)
-stream = JuliaSyntax.ParseStream(text)
-JuliaSyntax.parse!(stream)
+let filename = joinpath(@__DIR__, "literal_parsing.jl")
+    text = read(filename, String)
+    stream = JuliaSyntax.ParseStream(text)
+    JuliaSyntax.parse!(stream)
+end
--- a/src/JuliaSyntax.jl
+++ b/src/JuliaSyntax.jl
@@ -26,5 +26,5 @@ include("expr.jl")
 
 # Hooks to integrate the parser with Base
 include("hooks.jl")
-
+include("precompile.jl")
 end


src/precompile.jl
```diff
@@ -0,0 +1,5 @@
+# Just parse some file as a precompile workload
+filename = joinpath(@__DIR__, "literal_parsing.jl")
+text = read(filename, String)
+stream = JuliaSyntax.ParseStream(text)
+JuliaSyntax.parse!(stream)
```
aviatesk, line 5:
```suggestion
let filename = joinpath(@__DIR__, "literal_parsing.jl")
    text = read(filename, String)
    stream = JuliaSyntax.ParseStream(text)
    JuliaSyntax.parse!(stream)
end
```

KristofferC: Ah, of course. Thanks.

timholy:
I've been wondering about this too. Cthulhu now precompiles the whole chain, so in a sense that takes care of it for my needs. But having precompilation here should reduce the amount of time spent precompiling Cthulhu. We might want to precompile even more, so this seems like a good start (no need for you to do it).

It would be worth contemplating whether we want to switch to SnoopPrecompile; https://discourse.julialang.org/t/psa-for-snoopprecompile-turning-off-extra-workload-for-specific-packages/92865/5 makes it a bit more convenient for developers than having a "manual" approach. But of course, if this is going to end up in the sysimg, it may not be necessary.

c42f (approved):
Looks good to me.

What I've been doing previously when compiling a custom sysimage is running several of the tests:

https://github.com/JuliaLang/JuliaSyntax.jl/blob/main/sysimage/precompile.jl

But triggering more precompilation as part of the normal package top level code seems useful.

c42f:
> SnoopPrecompile is only needed to scoop up runtime-dispatched call graphs

Yes I think the dispatch here should all be pretty static. But I should probably check again sometime soon. There's been a lot of changes since last time I ran JET against this code.

timholy:
Yep, you're fine:
```julia
julia> using JuliaSyntax, SnoopCompile

julia> str = """
       function summer(list)
           s = 0
           for x in list
               s += x
           end
           return s
       end
       """
"function summer(list)\n    s = 0\n    for x in list\n        s += x\n    end\n    return s\nend\n"

julia> tinf = @snoopi_deep JuliaSyntax.parse(JuliaSyntax.SyntaxNode, str);

julia> itrigs = inference_triggers(tinf)
1-element Vector{InferenceTrigger}:
 Inference triggered to call JuliaSyntax.parse(::Type{JuliaSyntax.TreeNode{JuliaSyntax.SyntaxData}}, ::String) from eval (./boot.jl:370) inlined into REPL.eval_user_input(::Any, ::REPL.REPLBackend, ::Module) (/cache/build/default-amdci5-5/julialang/julia-release-1-dot-9/usr/share/julia/stdlib/v1.9/REPL/src/REPL.jl:153)
```
This doesn't show there's no runtime dispatch, but it does show that there were no such dispatches that required additional type inference. From the standpoint of precompilation, that's all that really matters anyway.

c42f:
Thanks Tim!

KristofferC:
OhMyREPL now runs with JuliaSyntax: https://github.com/KristofferC/OhMyREPL.jl/pull/277 :)

c42f:
Nice!

Issue #150, opened 2022-11-18
[Idea] a couple of compiler parser projects


camilogarciabotero:
Hi,

This is not really an issue, but a couple of projects I found could be helpful and inspire features. I came across this https://github.com/zesterer/chumsky and this https://github.com/zesterer/ariadne Rust projects for parsing compiler errors and got nice diagnostic renderings, I don’t really know if there is an overlap with the JuliaSyntax.jl project, but thought might spark others ideas :smiley: 

![image](https://user-images.githubusercontent.com/25586362/202769318-be950d8b-30d8-42cb-85f5-2104cb420e54.png)

PD: I was encouraged to post this from a Slack channel.

c42f:
Wow what a beautiful error message!

If someone wants to work on beautifying the error message rendering that would be great. I will probably get to it eventually :-)

FedeClaudi:
Hello everyone, I'm the developer of [Term.jl](https://github.com/FedeClaudi/Term.jl). I just wanted to say, this issue inspired me to add this kind of "annotations" in the image above to Term as a new feature. This is what it looks like:
```julia
Annotation(
     highlight("var::Int = 2x + \"a\""), 
    "2x + \"a\"", 
     "This seems like a bad idea"; 
     style="bright_yellow"
) |> print   
```
<img width="400" alt="image" src="https://user-images.githubusercontent.com/17436313/208676446-19f2a961-cd33-4c0b-87f8-284c72d087e9.png">

If it's something you'd find useful, let me know!

c42f:
@FedeClaudi this looks beautiful :heart: (Sorry didn't mean to ignore you, I just missed your message.)

If you'd like to help develop code in JuliaSyntax for rendering error messages that would be amazing! The key thing we need in this package is highlighting and annotating byte ranges within a file. You can see how this currently works here:

https://github.com/JuliaLang/JuliaSyntax.jl/blob/45d843186a335c63c9b13993e5352ae17d6382ec/src/diagnostics.jl#L44

---

Dependencies are a tricky thing for JuliaSyntax.jl as we need to vendor it into Base and eventually make it part of bootstrap somehow. So I wonder how we'd deal with that. Depending on Term.jl could be a problem if we have to vendor Term and all its dependencies into Base as well.

FedeClaudi:
[...]

PR #215, opened 2023-03-10
Rewrite source range `highlight()`

c42f:
`highlight()` now always prints whole lines of source code, and can highlight arbitrary ranges using box drawing characters, not just with ANSI colors.

Use this to fix diagnostic printing so that it's comprehensible in a non color terminal and so that pasting errors into non-color environments works, as described in https://github.com/JuliaLang/JuliaSyntax.jl/issues/150#issuecomment-1452960722.

The box drawing characters from WGL4 seem like a good balance of:

* Relatively compatible because they're very old, dating from DOS era: https://en.wikipedia.org/wiki/Box-drawing_character#DOS - note we also use these for formatting log messages from Logging.ConsoleLogger.

* Easy to distinguish from the user's source code

I've also found it's helpful to prepend any lines of annotation with a Julia # comment where possible - this ensures that copy+paste into websites with syntax highlighting will highlight the annotations separately from the code.

A simple example:

```julia
julia> (x - (c <--- d))
ERROR: ParseError:
(x - (c <--- d))
#       └──┘ ── invalid operator @ REPL[48]:1:9
```

With color

line_printing_2

Also some semi-related changes included

* Generalized/expanded _printstyled() function
* Better diagnostic range for try-without-catch

### DIFF
--- a/src/diagnostics.jl
+++ b/src/diagnostics.jl
@@ -40,6 +40,18 @@ end
 first_byte(d::Diagnostic) = d.first_byte
 last_byte(d::Diagnostic)  = d.last_byte
 is_error(d::Diagnostic)   = d.level == :error
+Base.range(d::Diagnostic) = first_byte(d):last_byte(d)
+
+# Make relative path into a file URL
+function _file_url(filename)
+    @static if Sys.iswindows()
+        # TODO: Test this with windows terminal
+        path = join(splitpath(abspath(filename)), '/')
+    else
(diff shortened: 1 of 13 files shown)


c42f:
Ah you got to be kidding. `splitpath()` didn't exist in julia-1.0? Ok, `replace()` it is I guess...

c42f:
codecov keeping me honest here :sweat_smile:

PR #233, opened 2023-03-27
fix lexing `'\\` to not error in the validate token step

### LINKED ISSUE #232: `EOFError: read end of file` when tokenizing a non finished char literal with backslash
```julia
julia> JuliaSyntax.tokenize("'\\")
ERROR: EOFError: read end of file
Stacktrace:
 [1] read
   @ ./iobuffer.jl:217 [inlined]
 [2] read
   @ ./io.jl:789 [inlined]
 [3] validate_tokens(stream::JuliaSyntax.ParseStream)
   @ JuliaSyntax ~/.julia/packages/JuliaSyntax/EGQXe/src/parse_stream.jl:903
 [4] #parse!#81
   @ ~/.julia/packages/JuliaSyntax/EGQXe/src/parser_api.jl:54
 [5] parse!
   @ ~/.julia/packages/JuliaSyntax/EGQXe/src/parser_api.jl:43 [inlined]
 [6] tokenize(text::String)
   @ JuliaSyntax ~/.julia/packages/JuliaSyntax/EGQXe/src/parser_api.jl:181
 [7] top-level scope
   @ REPL[4]:1
```

Reported in https://github.com/KristofferC/OhMyREPL.jl/issues/314. OhMyREPL assumes that the Tokenize step does not error so it is a regression there vs Tokenize.jl which produces:

```julia
julia> collect(Tokenize.tokenize("'\\"))
2-element Vector{Tokenize.Tokens.Token}:
 1,1-1,2          ERROR          "'\\"
 1,3-1,2          ENDMARKER      ""
```

KristofferC:
Fixes #232

### DIFF
--- a/src/literal_parsing.jl
+++ b/src/literal_parsing.jl
@@ -243,6 +243,9 @@ function unescape_julia_string(io::IO, str::AbstractString,
         escstart = i
         i += 1
         if i >= endind
+            emit_diagnostic(diagnostics, escstart:endind-1,
+                            error="invalid escape sequence")
+            had_error = true
             break
         end
         c = str[i]
--- a/src/parse_stream.jl
+++ b/src/parse_stream.jl
@@ -918,15 +918,11 @@ function validate_tokens(stream::ParseStream)
                 error_kind = K"ErrorInvalidEscapeSequence"
             else
                 seek(charbuf,0)
-                if eof(charbuf)
-                    error_kind = K"ErrorInvalidEscapeSequence" # for example "'\\"
-                else
-                    read(charbuf, Char)
-                    if !eof(charbuf)
-                        error_kind = K"ErrorOverLongCharacter"
-                        emit_diagnostic(stream, fbyte:lbyte,
-                                        error="character literal contains multiple characters")
-                    end
+                read(charbuf, Char)
+                if !eof(charbuf)
+                    error_kind = K"ErrorOverLongCharacter"
+                    emit_diagnostic(stream, fbyte:lbyte,
+                                    error="character literal contains multiple characters")
                 end
             end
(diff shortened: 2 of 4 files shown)


c42f (approved):
Looks good to me!

KristofferC:
Do you think the error I chose here is ok? I guess technically, it is not an invalid escape sequence?

c42f:
Yes I think that error is fine - it can be seen as an incomplete escape sequence.

Though now that I look closer, I think the fix is better made inside `unescape_julia_string`.  Let me push that variant up here.

pfitzseb:
Maybe introducing a new error kind (`ErrorEOF` or something like that) makes sense? Could then use that for other incomplete tokens as well, instead of the generic `K"error"`:
```
julia> JuliaSyntax.tokenize(""""foo""")
3-element Vector{JuliaSyntax.Token}:
 JuliaSyntax.Token(JuliaSyntax.SyntaxHead(K""", 0x0001), 0x00000001:0x00000001)
 JuliaSyntax.Token(JuliaSyntax.SyntaxHead(K"String", 0x0000), 0x00000002:0x00000004)
 JuliaSyntax.Token(JuliaSyntax.SyntaxHead(K"error", 0x0001), 0x00000005:0x00000004)
```

c42f:
There, I've pushed the tweak up into `unescape_julia_string` - that way we get the correct diagnostic emitted for strings as well as character literals.

c42f:
If you feel like making a release that's cool (just cherry pick this onto the `release-0.3.3` branch ... which uh.. should have been named `release-0.3`. Because `main` branch has breaking changes right now).

If not I'll do it tomorrow morning.

PR #238, opened 2023-03-31
Conservative exports + rename parse() -> parsex()

c42f:
`JuliaSyntax.parse()` clashes unfortunately with `Base.parse()`. Choose `parsex` for "parse expression" as a replacement.  Actually I think this might be good anyway, because it's really not clear to me whether `parse()` should parse a single expression or a whole file top-level.  Having a less generic name helps with that I think.

Also add a conservative list of exports that I expect "people are likely to use". There's more to the API than this, but shoving that all into the user's namespace doesn't seem ideal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ the `call` has the infix `-i` flag):
 ```julia
 julia> using JuliaSyntax
 
-julia> parsex(SyntaxNode, "(x + y)*z", filename="foo.jl")
+julia> parsestmt(SyntaxNode, "(x + y)*z", filename="foo.jl")
 line:col│ tree                                   │ file_name
    1:1  │[call-i]                                │foo.jl
    1:1  │  [parens]
@@ -72,7 +72,7 @@ representation, despite being important for parsing.
 
 ```julia
 julia> text = "(x + y)*z"
-       greentree = parsex(JuliaSyntax.GreenNode, text)
+       greentree = parsestmt(JuliaSyntax.GreenNode, text)
      1:9      │[call]
      1:7      │  [parens]
      1:1      │    (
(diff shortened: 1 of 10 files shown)


c42f:
@timholy I'd value your thoughts on this if you have time.

I've been putting off choosing an export list, but it's nice to have some basic functionality included with `using JuliaSyntax`.  So here's a stab at that.

KristofferC:
> `JuliaSyntax.parse()` clashes unfortunately with `Base.parse()`. 

In that sense, `Meta.parse` and `TOML.parse` clashes with `Base.parse`. It hasn't really been a problem though.

c42f:
Thinking a bit more about this, there's some terminological weirdness in Julia surrounding "statement" vs "expression"
* *Every* Julia syntactic construct is an expression in the sense that it yields a value and Julia code can be written in quite a functional style as a result.
* But typical Julia code uses side effects and is written in an imperative style as a sequence of statements (effectively disregarding the fact that each statement is also an expression).

So it's best to avoid the term "expression" in the naming of these parsing functions as it's extremely ambiguous. That really leaves us with `parsestmt()`, `parsestatement()`, `parse_stmt()`, `parse_statement()`.

`parse_stmt()` is a good suggestion but there's an additional "constraint" here - we already have `Meta.parseall()` and `Meta.parseatom()` in `Base` (without underscores) and we did debate those names in the past, so I'd like `JuliaSyntax.parseall()` and `JuliaSyntax.parseatom()` for consistency.

So I'm thinking the most consistent choices here are
* `parseall()`, `parsestmt()`, `parseatom()`
* `parseall()`, `parsestatement()`, `parseatom()`

c42f:
Ok, I made the tweak `parsestmt()` and changed `rule=:toplevel` to `rule=:all` in `parse!()` and added some depwarns for the next release so I think I'm happy with this now.

PR #244, opened 2023-04-19
Use importpath kind rather than K"."

c42f:
In `Expr` trees, the internal structure of children of `Expr(:.)` is different within an import statement vs outside. For example, `x.y` parses as
* `(. x (quote y))` in normal code
* `(. x y)` inside `import` and `using`.

This causes awkwardness when writing expression manipulation code because import paths
* Need to be distinguished from normal `.` but
* Can be nested several layers deep in `:` and `as` nodes within an `import` or `using` parent node.

To avoid this situation, here I've created a new `K"importpath"` kind specifically for import paths.

Part of #88 I guess


c42f:
@BenChung maybe this one helps a bit with one small issue you may have faced in SemanticAST?

BenChung:
Yes, thanks - it'll be a bit before I can update SemanticAST, however.

c42f:
Sure, no need to use this immediately or anything! I was just curious about whether it was a case you'd found annoying :-)

PR #308, opened 2023-06-09
Syntax experiment: String format syntax eg `"$(x, digits=2)"`

c42f:
See https://github.com/JuliaLang/julia/issues/18703

```julia
julia> JuliaSyntax.parsestmt(SyntaxNode, "\"prefix \$(1.0, digits=2) suffix\"")
   1:1  │[string]
   1:2  │  "prefix "
   1:10 │  [format]
   1:11 │    1.0
   1:16 │    [=]
   1:16 │      digits
   1:23 │      2
   1:25 │  " suffix"


julia> JuliaSyntax.parsestmt(Expr, "\"prefix \$(1.0, digits=2) suffix\"")
:("prefix $($(Expr(:format, 1.0, :(digits = 2)))) suffix")
```

CC @simonbyrne

### DIFF
--- a/src/kinds.jl
+++ b/src/kinds.jl
@@ -887,6 +887,7 @@ const _kind_names =
         "inert"          # QuoteNode; not quasiquote
         "juxtapose"      # Numeric juxtaposition like 2x
         "string"         # A string interior node (possibly containing interpolations)
+        "format"         # String formatting specification
         "cmdstring"      # A cmd string node (containing delimiters plus string)
         "char"           # A char string node (containing delims + char data)
         "macrocall"
--- a/src/parser.jl
+++ b/src/parser.jl
@@ -3182,14 +3182,16 @@ function parse_string(ps::ParseState, raw::Bool)
                 m = position(ps)
                 bump(ps, TRIVIA_FLAG)
                 opts = parse_brackets(ps, K")") do had_commas, had_splat, num_semis, num_subexprs
-                    return (needs_parameters=false,
+                    return (needs_parameters=true,
                             simple_interp=!had_commas && num_semis == 0 && num_subexprs == 1)
                 end
                 if !opts.simple_interp || peek_behind(ps, skip_parens=false).kind == K"generator"
                     # "$(x,y)" ==> (string (parens (error x y)))
-                    emit(ps, m, K"error", error="invalid interpolation syntax")
+                    emit(ps, m, K"format")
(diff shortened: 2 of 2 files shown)


giordano:
Can we use semicolons to start keyword arguments here?

c42f:
> Can we use semicolons to start keyword arguments here?

Certainly! But is it a good idea to diverge from the syntax for keyword args supported everywhere else in the language?

giordano:
> But is it a good idea to diverge from the syntax for keyword args supported everywhere else in the language?

 Uhm? Keyword arguments [do start with semicolons](https://docs.julialang.org/en/v1/manual/functions/#Keyword-Arguments) in the language already, sadly that's optional and using commas can lead to [not-so-fun-to-debug problems](https://discourse.julialang.org/t/keyword-behaves-differently-using-or/78233).

c42f:
> Keyword arguments [do start with semicolons](https://docs.julialang.org/en/v1/manual/functions/#Keyword-Arguments) in the language already

Haha! I do know about this. And some other fiddly details of Julia's syntax ;-)

My point is, I think it'd be confusing not to support the comma notation, given that it's supported and extremely widely used for normal function calls.  I think consistency is more important here than trying to fix problems which can't be fixed in the other more widely used syntax due to compat concerns.

andyferris:
I also kinda feel like semicolons might be good here, but it's not a super strong opinion. For one thing it wasn't obvious to me (before I checked) that `"$(1,2)` isn't interpolating the tuple `(1,2)`.

> can't be fixed in the other more widely used syntax due to compat concerns

This isn't the right forum/issue - but I rather like the idea of having something like Rust's "editions" to make breaking syntax improvements over time (without needing e.g. Julia 2.0), hence I'd challenge what "can't" be done (eventually...).

c42f:
> I rather like the idea of having something like Rust's "editions" to make breaking syntax improvements over time (without needing e.g. Julia 2.0), hence I'd challenge what "can't" be done (eventually...).

I agree with this and I've been thinking a lot about it for a while :-) I don't know exactly how it should work yet but I've been wanting to write a julep on it :)

c42f:
> I rather like the idea of having something like Rust's "editions" to make breaking syntax improvements over time

See

https://github.com/JuliaLang/julia/issues/54903

c42f:
Note about moving this to base, if necessary

```bash
# cd julia_dir

git fetch JuliaSyntax
git checkout pr-for-Base/308
git rebase origin/master
```

PR #322, opened 2023-07-01
AST: Rearrange `do` to sit inside `call`/`macrocall`

### LINKED ISSUE #319: What to do about "`do` macrocalls"
"`do` macrocalls" have an expression representation which is the same as `do` function calls.  But the macrocall in this case acts *not only* on the its children, but on the body of the `do` block.

It seems pretty weird for a macro expansion to act outside its child list in the expression tree, and this is the only such case.

Currently we have the following representation

```
julia> parsestmt(SyntaxNode, "@f(x, y) do a, b\n body\n end")
   1:1  │[do]
   1:1  │  [macrocall-p]
   1:2  │    @f
   1:4  │    x
   1:7  │    y
   1:12 │  [tuple]
   1:13 │    a
   1:16 │    b
   1:17 │  [block]
   2:2  │    body
```

Is there a neat alternative AST we could use to fix this special case?

c42f:
`do` syntax is represented in `Expr` with the `do` outside the call. This makes some sense syntactically (do appears as "an operator" after the function call).

However semantically this nesting is awkward because the lambda represented by the do block is passed to the call. This same problem occurs for the macro form `@f(x) do \n body end` where the macro expander needs a special rule to expand nestings of the form `Expr(:do, Expr(:macrocall ...), ...)`, rearranging the expression which are passed to this macro call rather than passing the expressions up the tree.

In this PR, we change the parsing of

    @f(x, y) do a, b\n body\n end
    f(x, y) do a, b\n body\n end

to tack the `do` onto the end of the call argument list:

    (macrocall @f x y (do (tuple a b) body))
    (call f x y (do (tuple a b) body))

This achieves the following desirable properties
1. Content of `do` is nested inside the call which improves the match between AST and semantics
2. Macro can be passed the syntax as-is rather than the macro expander rearranging syntax before passing it to the macro
3. In the future, a macro can detect when it's being passed do syntax rather than lambda syntax
4. `do` head is used uniformly for both call and macrocall
5. We preserve the source ordering properties we need for the green tree.

Fix #319

### DIFF
--- a/docs/src/reference.md
+++ b/docs/src/reference.md
@@ -43,7 +43,7 @@ the source text more closely.
 * The right hand side of `x where {T}` retains the `K"braces"` node around the `T` to distinguish it from `x where T`.
 * Ternary syntax is not immediately lowered to an `if` node: `a ? b : c` parses as `(? a b c)` rather than `Expr(:if, :a, :b, :c)` (#85)
 * `global const` and `const global` are not normalized by the parser. This is done in `Expr` conversion (#130)
-* The AST for `do` is flatter and not lowered to a lambda by the parser: `f(x) do y ; body end` is parsed as `(do (call f x) (tuple y) (block body))` (#98)
+* [`do` syntax](#Do-blocks) is nested as the last child of the call which the `do` lambda will be passed to (#98, #322)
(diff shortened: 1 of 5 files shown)


c42f:
@BenChung you might also appreciate this one (I hope :-) )

PR #326, opened 2023-07-10
Fix code block indentation

savq:
The code block has to be indented at the same level as the list elements, otherwise Documenter won't parse it correctly.

Screenshot 2023-07-10 at 16-31-36 Design Discussion · JuliaSyntax jl


c42f:
Sure, thanks!

PR #337, opened 2023-08-02
Make `SourceFile.code` infer concretely

timholy:
For example, without this, iteration over `.code` yields `AbstracChar`s rather than `Char`s.


aviatesk (approved):

c42f:
Thanks for the fix, having this abstract was not intentional!

At some point I want to redo `SourceFile` completely maybe along the lines of #190.

PR #397, opened 2023-12-05
Improve error message for missing closing tokens

### LINKED ISSUE #313: better error message for missing comma.
When missing a `,` in a list or function args, the error message currently says "expected `)`" when the real problem is the missing comma. Should the error message be updated to add that as a possibility? (examples where this comes up are `+(1, 2 3)` and `[1, 2 3]`)

### LINKED ISSUE #349: Underbracket placement
```jl
ERROR: ParseError:
# Error @ REPL[53]:15:5
    ylims!(p, (0, last(ylims(p)))
    xlabel!(p, "Contig length cutoff (kbp)")
#   └─────────────────────────────────────┘ ── Expected `)`
```

[On Slack](https://julialang.slack.com/archives/C6FGJ8REC/p1693495800335689?thread_ts=1693472893.530149&cid=C6FGJ8REC) it was pointed out that the underbracket ends unexpectedly early. If the last place a `)` could appear is after the final `)` then that's where the `┘` should be too -- right after the final `)`.

c42f:
When a missing closing token like `)`, `]` or `}` is encountered we want the "Expected `)`" error to point to a location one past the last valid token, not to the trailing error tokens.

For example from #349 here's a poor error message from the existing code:

    ERROR: ParseError:
    # Error @ REPL[53]:15:5
        ylims!(p, (0, last(ylims(p)))
        xlabel!(p, "Contig length cutoff (kbp)")
    #   └─────────────────────────────────────┘ ── Expected `)`

After this change, the error location instead points to the end of the last valid line:

    ERROR: ParseError:
    # Error @ REPL[53]:15:5
        ylims!(p, (0, last(ylims(p)))
    #                                └── Expected `)`

Fix #349, fix #313

### DIFF
--- a/src/parser.jl
+++ b/src/parser.jl
@@ -141,15 +141,20 @@ end
 #
 # Crude recovery heuristic: bump any tokens which aren't block or bracket
 # closing tokens.
-function bump_closing_token(ps, closing_kind)
+function bump_closing_token(ps, closing_kind, alternative_closer_hint=nothing)
     # todo: Refactor with recover() ?
-    bump_trivia(ps)
     if peek(ps) == closing_kind
+        bump_trivia(ps)
         bump(ps, TRIVIA_FLAG)
         return
     end
+    errmsg = "Expected `$(untokenize(closing_kind))`"
+    if !isnothing(alternative_closer_hint)
+        errmsg *= alternative_closer_hint
+    end
     # We didn't find the closing token. Read ahead in the stream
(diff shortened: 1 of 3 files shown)


tecosaur:
A thought on this: when the tokes are paired, it could be helpful to highlight the corresponding un-paired token.

For example, here's some (lisp) code in my editor when I delete an opening paren:

image

PR #420, opened 2024-02-25
Remove unused `is_error` import in Tokenize module

ericphanson:
I was testing [ExplicitImports.jl](https://github.com/ericphanson/ExplicitImports.jl) and noticed this unnecessary import:
````julia
julia> using JuliaSyntax, ExplicitImports

julia> print_explicit_imports(JuliaSyntax)
Module JuliaSyntax is not relying on any implicit imports.
Module JuliaSyntax.Tokenize is not relying on any implicit imports.

Additionally JuliaSyntax.Tokenize has stale explicit imports for these unused names:
JuliaSyntax
is_error
````

I left the `JuliaSyntax` one since I figured it could be annoying if you are working in that module and the name is unexpectedly unavailable. I feel that removing the `is_error` one improves clarity though, as it makes it easier to see that Tokenize does not add a method for `is_error` (or even use it).

### DIFF
--- a/src/tokenize.jl
+++ b/src/tokenize.jl
@@ -5,7 +5,7 @@ export tokenize, untokenize
 using ..JuliaSyntax: JuliaSyntax, Kind, @K_str, @KSet_str
 
 import ..JuliaSyntax: kind,
-    is_literal, is_error, is_contextual_keyword, is_word_operator
+    is_literal, is_contextual_keyword, is_word_operator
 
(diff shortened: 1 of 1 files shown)


ericphanson:
I think the CI failures are unrelated (since I wouldn't expect the import is needed on some Julia versions and not others), though I don't really understand the failures so I'm not 100% sure.

c42f (approved):

PR #421, opened 2024-02-25
add more specific error for invalid identifier start chars

JeffBezanson:
Some characters can continue but not start an identifier. Currently these give an "unknown unicode character" error, which could be more accurate.


c42f (approved):
Sounds good. I added an extra test and tweaked the error message to include the character in analogy to the existing ErrorUnknownCharacter etc + rebased to get the CI fixes.

Issue #416, opened 2024-02-17
Throw on invalid keywords


LilithHafner:
Running `parseall(SyntaxNode, "abc", ignore_error=true)` should throw because "ignore_error" is not a valid keyword.

```julia
julia> parseall(SyntaxNode, "abc", ignore_error=true)
line:col│ tree                                   │ file_name
   1:1  │[toplevel]
   1:1  │  abc


julia> parseall(SyntaxNode, "abc +", ignore_error=true)
ERROR: ParseError:
# Error @ line 1:6
abc +
#    └ ── premature end of input
Stacktrace:
 [1] _parse(rule::Symbol, need_eof::Bool, ::Type{…}, text::String, index::Int64; version::VersionNumber, ignore_trivia::Bool, filename::Nothing, first_line::Int64, ignore_errors::Bool, ignore_warnings::Bool, kws::@Kwargs{…})
   @ JuliaSyntax ~/.julia/packages/JuliaSyntax/eqCSU/src/parser_api.jl:93
 [2] _parse (repeats 2 times)
   @ ~/.julia/packages/JuliaSyntax/eqCSU/src/parser_api.jl:77 [inlined]
 [3] #parseall#96
   @ ~/.julia/packages/JuliaSyntax/eqCSU/src/parser_api.jl:143 [inlined]
 [4] top-level scope
   @ REPL[306]:1
Some type information was truncated. Use `show(err)` to see complete types.

julia> parseall(SyntaxNode, "abc +", ignore_errors=true)
line:col│ tree                                   │ file_name
   1:1  │[toplevel]
   1:1  │  [call-i]
   1:1  │    abc
   1:5  │    +
   1:6  │    ✘
```

Issue #341, opened 2023-08-07
MethodError: no method matching `isless(::Int64, ::Nothing)`


pchintalapudi:
MWE:

```julia
premc@DESKTOP-0UND5EL:/mnt/c/Users/premc/Documents/julia$ ./julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.11.0-DEV.235 (2023-08-06)
 _/ |\__'_|_|_|\__'_|  |  Commit 9f9e989f24 (1 day old master)
|__/                   |

julia> ./julia
┌ Error: JuliaSyntax parser failed — falling back to flisp!
│   exception =
│    MethodError: no method matching isless(::Int64, ::Nothing)
│ 
│    Closest candidates are:
│      isless(::Missing, ::Any)
│       @ Base missing.jl:87
│      isless(::Any, ::Missing)
│       @ Base missing.jl:88
│      isless(::Real, ::AbstractFloat)
│       @ Base operators.jl:178
│      ...
│ 
│    Stacktrace:
│      [1] <(x::Int64, y::Nothing)
│        @ Base ./operators.jl:352
│      [2] >(x::Nothing, y::Int64)
│        @ Base ./operators.jl:378
│      [3] core_parser_hook(code::String, filename::String, lineno::Int64, offset::Int64, options::Symbol)
│        @ Base.JuliaSyntax /mnt/c/Users/premc/Documents/julia/base/JuliaSyntax/src/hooks.jl:212
│      [4] invoke_in_world(::UInt64, ::Any, ::Any, ::Vararg{Any}; kwargs::@Kwargs{})
│        @ Base ./essentials.jl:921
│      [5] invoke_in_world(::UInt64, ::Any, ::Any, ::Vararg{Any})
│        @ Base ./essentials.jl:918
│      [6] (::Base.JuliaSyntax.var"#invoke_fixedworld#120"{Base.JuliaSyntax.var"#invoke_fixedworld#117#121"{typeof(Base.JuliaSyntax.core_parser_hook), UInt64}})(::String, ::Vararg{Any}; kws::@Kwargs{})
[...]

PR #422, opened 2024-02-25
unhandled kwargs throw errors

ericphanson:
closes https://github.com/JuliaLang/JuliaSyntax.jl/issues/416

My editor is configured to remove trailing whitespace, which caused some diffs here. Let me know if that's too annoying and I'll undo it.

Locally I get the test failures:

```julia
Parse core.jl: Test Failed at /Users/eph/JuliaSyntax.jl/test/test_utils.jl:237
  Expression: parsers_agree

Parse core.jl: Test Failed at /Users/eph/JuliaSyntax.jl/test/test_utils.jl:241
  Expression: reduced_failures == []
   Evaluated: AbstractString["for f() in 1:3\n        push!(a, f())\n        push!(fs, f)\n    end", " {{}}where{}"] == Any[]
```

same as in #420, which I think are unrelated.

### DIFF
--- a/src/green_tree.jl
+++ b/src/green_tree.jl
@@ -87,11 +87,11 @@ function Base.show(io::IO, ::MIME"text/plain", node::GreenNode, str::AbstractStr
     _show_green_node(io, node, "", 1, str, show_trivia)
 end
 
-function build_tree(::Type{GreenNode}, stream::ParseStream; kws...)
-    build_tree(GreenNode{SyntaxHead}, stream; kws...) do h, srcrange, cs
+# we don't need the `filename` and `first_line` kwargs, but some trees do, so we allow them to be passed
+function build_tree(::Type{GreenNode}, stream::ParseStream; filename=nothing, first_line=nothing)
+    build_tree(GreenNode{SyntaxHead}, stream) do h, srcrange, cs
         span = length(srcrange)
         isnothing(cs) ? GreenNode(h, span, ()) :
                         GreenNode(h, span, collect(GreenNode{SyntaxHead}, cs))
     end
 end
-
--- a/src/expr.jl
+++ b/src/expr.jl
@@ -140,7 +140,7 @@ function _string_to_Expr(k, args)
         #   """\n  a\n  b""" ==>  "a\nb"
         return only(args2)
     else
-        # This only happens when k == K"string" or when an error has occurred. 
+        # This only happens when k == K"string" or when an error has occurred.
         return Expr(:string, args2...)
     end
 end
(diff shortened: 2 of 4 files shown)


src/green_tree.jl
```diff
@@ -87,11 +87,11 @@ function Base.show(io::IO, ::MIME"text/plain", node::GreenNode, str::AbstractStr
     _show_green_node(io, node, "", 1, str, show_trivia)
 end
 
-function build_tree(::Type{GreenNode}, stream::ParseStream; kws...)
-    build_tree(GreenNode{SyntaxHead}, stream; kws...) do h, srcrange, cs
+# we don't need the `filename` and `first_line` kwargs, but some trees do, so we allow them to be passed
+function build_tree(::Type{GreenNode}, stream::ParseStream; filename=nothing, first_line=nothing)
```
c42f, line 91: It's confusing to know what to do with this part: do we want all `build_tree(::Type{TreeType}, ...)` methods to need to know about each other for all `TreeType`? This seems impossible (eg, external packages) which would imply these should just be an error rather than ignored.

But on the other hand making this an error means `build_tree()` with keywords can't be used generically.

It seems there's two consistent things to do:
* Be generic and ignore unknown keywords (current status quo. not a good user experience, as in #341)
* Throw on unknown keywords (makes `parseall` etc couple the `TreeType` with keywords)

This PR takes an awkward middle path between those two options. Which might be ok but seems inconsistent?

ericphanson: Yeah, I guess if we want a more decoupled approach, we could copy Makie and add a `validate_kwargs` function which is called on the tree type and the passed-in kwargs (and throws or returns `nothing`), and is called in `parseall` or `_parse` or something like that. Then custom trees which wish to support additional kwargs would need to define a `validate_kwargs` method to allow them to pass through.

c42f: That sounds promising! So we define `validate_parse_args(::Type{GreenNode}; kws...)` and users with custom trees can overload it?

c42f:
Heya, we just moved JuliaSyntax into Base in https://github.com/JuliaLang/julia/pull/59870 - if you want to re-propose this PR over there, I've created a branch https://github.com/JuliaLang/JuliaSyntax.jl/tree/pr-for-Base/422 to make this easier.

* Clone JuliaLang/julia
* Check out pr-for-Base/422
* Rebase on top of master

For example:

```bash
# cd julia_dir

git fetch JuliaSyntax
git checkout pr-for-Base/422
git rebase origin/master
```

PR #430, opened 2024-04-22
Delete unused down option from parse_docstring

LilithHafner:
This is an NFC simplification of control flow.

After this merges, I plan to follow up with swapping the order of parse_public and parse_docstring and then adjusting the quote callsites of parse_eq to be parse_public. This should fix #51450 without allowing docstrings in quotes. (or, if we'd like, we can also allow docstrings in quotes)


c42f (approved):
Sure. Idk why this is like it is. It seems historical.

PR #437, opened 2024-06-07
Allow typevars in `do` expressions

sgaure:
(@c42f edit:) Upstream syntax discussion: https://github.com/JuliaLang/julia/issues/54915

There are several ways to create functions in julia. One of them is the `do` syntax. You can't dispatch on `do` functions, but now and then the need comes up to capture the types of variables.  Although this is easy with `typeof` inside the function body, it turns out to be quite easy to enable the use of `where` clauses in the `do` syntax.  This would make a `do` function similar to the other function definitions.

The reason it's easy is that such expressions are already being parsed, but a minor detail prevents it from being accepted as a valid expression:
```
julia-1.10.4> Meta.show_sexpr(:(f() do (x::T) where {T}; T(x) end))
(:do, (:call, :f), (:->, (:tuple, (:where, (:(::), :x, :T), :T)), (:block,
      :(#= REPL[1]:1 =#),
      (:call, :T, :x)
    )))
```
The `where` clause becomes encapsulated in a `tuple`.  The reason is that after the parser sees `do`, it looks for a comma separated list. After the list is parsed, i.e. on `;` or linefeed, the list is encapsulated in a `tuple`. This means that the `where` clause becomes an argument to the function, which fails as a syntax error.  By modifying the parsing of `do` expressions to avoid inserting `tuple` if the newly parsed list is a `where`, the parse result becomes:
```
julia-this-PR> Meta.show_sexpr(:(f() do (x::T) where {T}; T(x) end))
(:do, (:call, :f), (:->, (:where, (:(::), :x, :T), :T), (:block,
      :(#= REPL[1]:1 =#),
      (:call, :T, :x)
    )))
```
This is a valid `:->` expression, just like an anonymous function definition `((x::T) where T) -> T(x)` currently is.

The PR is a one-line change to `parser.jl`, isolated to parsing what follows `do`. It does not enable any new or very useful functionality, it merely will make functions defined in `do` quite similar to other function definitions.

I don't think this change will break anything, as the construction currently results in a `ERROR: syntax: ...` failure.

### DIFF
--- a/src/parse_stream.jl
+++ b/src/parse_stream.jl
@@ -475,7 +475,7 @@ end
 end
 
 """
-    peek(stream [, n=1]; skip_newlines=false)
+    peek(stream::ParseStream [, n=1]; skip_newlines=false)
 
 Look ahead in the stream `n` tokens, returning the token kind. Comments and
 non-newline whitespace are skipped automatically. Whitespace containing a


LilithHafner:
I think it is best for proposed syntax changes to start as or have an associated issue in the JuilaLang/julia repo for viability.

test/parser.jl
```diff
@@ -369,7 +369,9 @@ tests = [
         "f() do x, y\n body end"  =>  "(call f (do (tuple x y) (block body)))"
         "f(x) do y body end"  =>  "(call f x (do (tuple y) (block body)))"
         "@f(x) do y body end" =>  "(macrocall-p @f x (do (tuple y) (block body)))"
-
+        "f(x) do (y::T) where T body end"  =>  "(call f x (do (where (parens (::-i y T)) T) (block body)))"
```
LilithHafner, line 372:
```suggestion
        "f(x) do y::T where T body end"  =>  "(call f x (do (tuple (::-i y (where T T))) (block body)))"
        "f(x) do (y::T) where T body end"  =>  "(call f x (do (where (parens (::-i y T)) T) (block body)))"
```
Maybe add a test for the case without parenthesis (IIUC the where clause will bind tighter than the ::).

LilithHafner (approved):
LGTM, technically.

LilithHafner:
See also: https://github.com/JuliaLang/julia/issues/55159

And also, as a syntax change this would need to be gated on VERSION.

c42f:
Upstream issue closed as not planned, so I'll close this for now.

If any of this code needs moving into Base we have the JuliaSyntax code there now.

Example of how to move it

```bash
# cd julia_dir

git fetch JuliaSyntax
git checkout pr-for-Base/437
git rebase origin/master
```

PR #440, opened 2024-06-16
Update `Base.peek` docstring

nhz2:
When looking at the help for `peek` in the repl, the current docstring looks like it applies to all `IO`, so I added `::ParseStream` to make the docstring more specific. See also https://github.com/JuliaLang/julia/issues/54749

### DIFF
--- a/src/parse_stream.jl
+++ b/src/parse_stream.jl
@@ -475,7 +475,7 @@ end
 end
 
 """
-    peek(stream [, n=1]; skip_newlines=false)
+    peek(stream::ParseStream [, n=1]; skip_newlines=false)
 
 Look ahead in the stream `n` tokens, returning the token kind. Comments and
 non-newline whitespace are skipped automatically. Whitespace containing a


GunnarFarneback:
This makes it clear that it's not an IO method of `peek` but most Base users won't have any idea what a `ParseStream` is and that they need to look it up as `Base.JuliaSyntax.ParseStream`.

But I think the real problem here is that JuliaSyntax shouldn't add a method to `Base.peek` at all but use a module local function for its internal ParseStream peeking.

KristofferC:
Yes, either we remove the docstring or make it module local. No need for this to show in docs at all.

pfitzseb:
Both of those seem like bad options, no? `Base.peek` is the right concept, so an overload seems appropriate. Removing the docstring seems weird as well.

Can't we do this level of filtering in Documenter.jl instead?

pfitzseb:
> why does it need a docstring?

Because it has additional kwargs that should be documented.

> Okay, how about detecting when the package is put into the sysimage and not add the docstring in that case?

Ok, so I think there are a few desirable properties for docstrings added by stdlibs:
1. Documentation can be retrieved when the stdlib is loaded (by any mechanism, sysimg or no)
2. stdlib docs are *distinct enough* from Base docs when viewed in the REPL (e.g. by printing the defining module)
3. Documenter should separate/filter docstrings based on where they're defined (so the Base section only talks about Base methods)

Most of that holds for generic packages as well, and this PR is probably not the right place to talk about that anyways.

---

I think I'll just go ahead and merge this as-is, because it's an improvement. Further steps can be taken in a new PR.

Issue #446, opened 2024-07-02
Better corpus testing


LilithHafner:
Whenever folks write new code in base or the stdlib that depends on JuliaSyntax bugfixes that have not been backported to flisp, CI will break on this repo. See #444, for an example where adding 

https://github.com/JuliaLang/julia/blame/f2558c461c85be4220901f4c67cbce718ddc015b/test/core.jl#L8115

In https://github.com/JuliaLang/julia/pull/52228 caused CI to fail here.

Short term solution: stop testing against test/core.jl
Long term solution: test against flisp using this corpus as of Julia v1.9 and do not include new code

LilithHafner:
I think that we should parse a massive corpus on the head and target commit of each PR and fail CI if there are any differences unless the PR is tagged with `syntax change`.

Issue #395, opened 2023-11-24
`A where {T} == Z` parsing inconsistency with flisp parser


c42f:
Tests are now failing on `main` due to a parsing inconsistency between JuliaSyntax and the flisp parser and a line in the JuliaLang/julia tests at

https://github.com/JuliaLang/julia/blob/6e23543bc477eb46e5fc8d5cab119190b990ed7c/test/core.jl#L8070

This can be reduced to an inconsistency parsing `A where {T} == Z`. One might expect this to parse with the `==` at precedence below the `where`, and indeed this is what happens in JuliaSyntax:

```julia
julia> JuliaSyntax.parsestmt(SyntaxNode, "A where {T} == Z")
line:col│ tree                                   │ file_name
   1:1  │[call-i]
   1:1  │  [where]
   1:1  │    A
   1:9  │    [braces]
   1:10 │      T
   1:13 │  ==
   1:16 │  Z
```

However, the flisp parser uses the following, apparently nonsensical precedence:

```julia
julia> dump(JuliaSyntax.fl_parse(Expr, "A where {T} == Z"))
Expr
  head: Symbol where
  args: Array{Any}((2,))
    1: Symbol A
    2: Expr
      head: Symbol call
      args: Array{Any}((3,))
        1: Symbol ==
        2: Expr
          head: Symbol braces
          args: Array{Any}((1,))
            1: Symbol T
        3: Symbol Z
```

The reason for this is that the flisp parser uses `parse-comparsion` for the trailing form because we want expressions without `{}` to allow subtype comparisons. That is, we want `A where T <: Z` to parse with the `where` at lower precedence than the `<:` as `(where A (<: T Z))`:

```julia
julia> JuliaSyntax.parsestmt(SyntaxNode, "A where T <: Z")
line:col│ tree                                   │ file_name
   1:1  │[where]
   1:1  │  A
   1:9  │  [<:]
   1:9  │    T
   1:14 │    Z
```

JuliaSyntax has a special case for trailing `{` for technical reasons which were originally unrelated to precedence. But it's arguably more useful and "correct" behavior.

So now we've got to figure out what to do about this. I think we should
1. Make the tests tolerant to this difference (short term fix for the tests)
2. Figure out a better, more stable way to do regression testing than always testing against the flisp parser and whatever code happens to be in Base on that version of Julia.

[...]

PR #447, opened 2024-07-02
Fix corpus tests on Julia 1.10+

### LINKED ISSUE #444: Tests fail on Julia 1.10 and 1.11-rc
I wanted to include running JuliaSyntax.jl tests into the VS Code extension release pipeline (https://github.com/julia-vscode/julia-vscode/pull/3625), but it turns out that tests fail on both Julia 1.10 and 1.11-rc. And it turns out that CI here isn't run for anything newer than 1.8. I think it would generally be good if this package here ran CI on minimally all released versions regularly, plus beta or release candidates.

LilithHafner:
Implements the short-term solution in #446, intented to fix #444 and implements step one in #395

### DIFF
--- a/test/parse_packages.jl
+++ b/test/parse_packages.jl
@@ -39,9 +39,12 @@ base_tests_path = joinpath(Sys.BINDIR, Base.DATAROOTDIR, "julia", "test")
             return nothing
         end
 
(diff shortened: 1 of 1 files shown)


LilithHafner:
Nightly tests still fail for unrelated reasons.

PR #448, opened 2024-07-02
Fix tests

LilithHafner:
- Don't assume (io::IOBuffer).data is a Vector on new versions of Julia
- Exclude test/subtype.jl from corpus due to new dependence on where precedence (#395)

### DIFF
--- a/src/parse_stream.jl
+++ b/src/parse_stream.jl
@@ -314,7 +314,8 @@ function ParseStream(text::AbstractString, index::Integer=1; version=VERSION)
 end
 
 # IO-based cases
-function ParseStream(io::IOBuffer; version=VERSION)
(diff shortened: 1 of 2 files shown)


src/parse_stream.jl
```diff
@@ -314,7 +314,8 @@ function ParseStream(text::AbstractString, index::Integer=1; version=VERSION)
 end
 
 # IO-based cases
-function ParseStream(io::IOBuffer; version=VERSION)
+# TODO: switch ParseStream to use a Memory internally on newer versions of Julia
+VERSION < v"1.11.0-DEV.753" && function ParseStream(io::IOBuffer; version=VERSION)
```
LilithHafner, line 318: This special case is not valid on new Julias where the internal io.data field is not a Vector.

c42f:
Thanks for fixing this :two_hearts:

PR #455, opened 2024-07-07
Fix emitting comments followed by `\r\n`

fredrikekre:
Before this patch the `\r` character would be part of the comment and not the following NewlineWs (which would simply be a `\n`-style NewlineWs instead of a `\r\n` as one would expect).

### DIFF
--- a/src/tokenize.jl
+++ b/src/tokenize.jl
@@ -747,8 +747,8 @@ function lex_comment(l::Lexer)
     if peekchar(l) != '='
         valid = true
         while true
-            pc = peekchar(l)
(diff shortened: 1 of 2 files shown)


LilithHafner:
In Julia 1.0 and 1.8 we have `Meta.parse("#1\r+2") === nothing`, would this PR change that?

LilithHafner:
Whether or not we merge this we should add a test for parsing `"#1\r+2"`

fredrikekre:
Fixed.

fredrikekre:
Whats the procedure to release a new version here?

Issue #457, opened 2024-07-11
`char_range()` function for indexing into source strings


davidanthoff:
We have a crash from the VS Code extension that seems to originate from JuliaSyntax. Repo steps are:
1. `pkg> dev Sunny`
2. Run
```julia
using JuliaSyntax

src = read(joinpath(homedir(), ".julia/dev/Sunny/test/test_tensors.jl"), String)

tree = parseall(SyntaxNode, src, filename="foo.jl")

i = JuliaSyntax.last_byte(tree[2][end][end])

src[i]
```

That crashes with
```
ERROR: StringIndexError: invalid index [2142], valid nearby indices [2140]=>'′', [2143]=>'\r'
```

I assume that `last_byte` should always return a valid index, right? So this seems like a bug.

The rest of this issue is some speculation from my end, might all be wrong :)

I started looking a bit around in the JuliaSyntax code, and there seems to be a fair bit of code that assumes 1-byte characters, which strikes me as incorrect? Or maybe this is carefully only done when there is a guarantee that only 1-byte codepoints can appear? An example is https://github.com/JuliaLang/JuliaSyntax.jl/blob/1d950817b8ab62a3be2464f4d5b097aa68c17042/src/parse_stream.jl#L519 Doesn't that (incorrectly) assume that the character at index position `last_byte` is only one code unit? Generally if I search the repo for `- 1` or `+ 1` I see a fair bit of code where I would have assumed that a `prevind` or `nextind` would be needed?

KristofferC:
Why would indexing into a string with the `last_byte` make sense? Seem the same conceptually as doing `"α"[sizeof("α")]`.

davidanthoff:
Yeah, I also realized, I should probably be using `range` instead of using `first_byte` and `last_byte`. At least for `range` I assume it should return valid indices, right? My PR would fix that, the repo for the problem with `range` is:
```julia
using JuliaSyntax

src = read(joinpath(homedir(), ".julia/dev/Sunny/test/test_tensors.jl"), String)

tree = parseall(SyntaxNode, src, filename="foo.jl")

code_range = range(tree[2][end][end])

src[code_range]
```
which also errors.

DaniGlez:
If anyone else is facing this issue, a way to bypass it until there's a permanent fix is to find your lines ending in non-ASCII Unicode characters and make that no longer the case.

MilesCranmer:
Thanks @DaniGlez!

[...]

PR #458, opened 2024-07-11
Ensure range returns valid indices

davidanthoff:
This (maybe) kind of fixes https://github.com/JuliaLang/JuliaSyntax.jl/issues/457. It at least ensures that `range` always is a valid range that can be used to index into the source string, i.e. it moves the logic that fixes the indices returned by `last_byte` from `Base.view` into `Base.range`.

Still wondering, though, whether the proper fix wouldn't be to do something about `span`.

### DIFF
--- a/src/syntax_tree.jl
+++ b/src/syntax_tree.jl
@@ -133,7 +133,7 @@ function sourcetext(node::AbstractSyntaxNode)
 end
 
 function Base.range(node::AbstractSyntaxNode)
-    (node.position-1) .+ (1:span(node))
+    node.position:prevind(node.source.code, node.position + span(node))
 end
 
 source_line(node::AbstractSyntaxNode) = source_line(node.source, node.position)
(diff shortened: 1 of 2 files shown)


src/syntax_tree.jl
```diff
@@ -133,7 +133,7 @@ function sourcetext(node::AbstractSyntaxNode)
 end
 
 function Base.range(node::AbstractSyntaxNode)
-    (node.position-1) .+ (1:span(node))
+    node.position:prevind(node.source.code, node.position + span(node))
```
c42f, line 136: It's intentional that `range()` specifically returns a byte range, so this part won't be correct without reviewing the other code.

(In fact, in a change I'm about to do, I'm renaming `range()` to `byte_range()`!)

davidanthoff: Ah, I see. I think it would be good to have some minimal docs about these things :)

Also, if it is supposed to  _not_ return string indices here, then there is probably some test missing that would actually fail with the changes in my PR here, right?

c42f: Mm more docs would be good.

At some point this package went from "a hacky thing I dream of being the main parser" to "oh wait it's actually important but still incomplete" :laughing:

PR #466, opened 2024-07-19
Use `K"function"` for short form function AST

### LINKED ISSUE #465: AST of short form functions
A pain point when writing macros is detecting all the types of things which might be lowered to functions and correctly extracting the formal arguments. This is due to
1. Existence of both short form and long form function definitions
2. The argument lists of anonymous functions being parsed into varying inconsistent `Expr` forms

JuliaSyntax already deals with part (2) of this.

I wonder if we should also deal with part (1) by emitting a clearer AST.

Currently we have

```julia
julia> parsestmt(SyntaxNode, "f()::T = 1")
   1:1  │[=]
   1:1  │  [::-i]
   1:1  │    [call]
   1:1  │      f
   1:6  │    T
   1:10 │  1
```

But it would be easy to change this into

```julia
julia> parsestmt(SyntaxNode, "f() = 1")
   1:1  │[function-=]
   1:1  │  [call]
   1:1  │    f
   1:7  │  1
```

And to have a syntax flag "`=`" set on the `function` node to indicate it was short-form syntax.

c42f:
Fix #465 - see rationale there.

Fixing this turned out to be quite easy.

Compat:

* This is a breaking change to the JuliaSyntax AST
* Non-breaking for `Expr`

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -118,6 +118,7 @@ JuliaSyntax.COLON_QUOTE
 JuliaSyntax.TOPLEVEL_SEMICOLONS_FLAG
 JuliaSyntax.MUTABLE_FLAG
 JuliaSyntax.BARE_MODULE_FLAG
+JuliaSyntax.SHORT_FORM_FUNCTION_FLAG
 ```
 
 ## Syntax trees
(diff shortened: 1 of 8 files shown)


domluna:
looks good to me. Does it handle all 4 of these?

```julia
f(x) = 10
f(x)::R = 10
f(x::T) where T = 10
f(x::T)::R where T = 10
```

c42f:
> Does it handle all 4 of these?

Yes!

PR #468, opened 2024-07-20
Backport release 0.4

fredrikekre:
I am sure there are more fixes to backport but these are the ones I care about right now. If I, or someone else, needs other fixes we can release 0.4.10 with more backports.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -39,6 +39,7 @@ jobs:
           # - 1.0
           # - 1.6
           # - 1
+          # - pre
           # - nightly
           # but remove some configurations from the build matrix to reduce CI time.
           # See https://github.com/marketplace/actions/setup-julia-environment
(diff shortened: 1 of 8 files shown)


fredrikekre:
@c42f do you know which commits need to be backported to make tests pass on 1.10? Looks like 0.4.8 also fail on 1.10 though... Perhaps we can merge and release this anyway?

c42f:
We'd need to backport most of @LilithHafner's recent CI fixes. I've done that, plus several other low risk items and pushed to the branch :)

c42f (approved):
I think this should be good now!

Everything here should be low risk fixes - do you want to merge this and make the release?

For big and higher risk changes I would also run the ecosystem parsing tests in `tools/check_all_packages.jl`.  But those aren't completely simple to set up (the package cache takes a while to download), and are somewhat nontrivial to interpret :grimacing:

fredrikekre:
Thanks.

PR #475, opened 2024-07-26
Newline macros

MasonProtter:
This is my current attempt at https://github.com/JuliaLang/JuliaSyntax.jl/issues/436, sorry it took me a little while to post it @c42f. 

Lots of things work, but the main problem currently is that my criterion for detecting cases like `(@foo x)` is too broad. Specifically, cases like this aren't doing what we want:
```
julia> """
       (@foo function bar()
           @baz x
           y
       end)
       """ |> Meta.parse
:(#= none:1 =# @foo function bar()
          #= none:1 =#
          #= none:2 =#
          #= none:2 =# @baz x y
      end)
```
Basically, it's detecting that `(@foo` is matching the desired pattern, but then its also recursively applying the new parsing rule to `@baz x` and grabbing the `y` incorrectly.  

I haven't yet figured out the right way to fix this.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,7 +4,7 @@ authors = ["Claire Foster <aka.c42f@gmail.com> and contributors"]
 version = "0.4.6"
 
 [compat]
-julia = "1.11"
+julia = "1.0"
 
 [deps]
 
--- a/Project.toml
+++ b/Project.toml
@@ -4,7 +4,7 @@ authors = ["Claire Foster <aka.c42f@gmail.com> and contributors"]
 version = "0.4.6"
 
 [compat]
-julia = "1.0"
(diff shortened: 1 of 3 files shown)


c42f:
Heya, we just moved JuliaSyntax into Base in https://github.com/JuliaLang/julia/pull/59870 - if you want to re-propose this PR over there, I've created a branch https://github.com/JuliaLang/JuliaSyntax.jl/tree/pr-for-Base/475 to make this easier.

* Clone JuliaLang/julia
* Check out pr-for-Base/475
* Rebase on top of master

For example:

```bash
# cd julia_dir

git fetch JuliaSyntax
git checkout pr-for-Base/475
git rebase origin/master
```

Issue #472, opened 2024-07-23
Release 1.0.0


c42f:
The next release will be 1.0.0 due to various breaking changes.

Most importantly, there are some breaking AST changes which were merged quite some time ago and we really need to get them into a released version as people are starting to really use this package for nontrivial tooling. Unfortunately, AST changes are kinda impossible to do depwarns for. I'm not sure what we can really do there other than making clear release notes. Also batching the breaking changes a bit so that packages which need to change can do those changes in one go?

My current plan for 1.0 is that I've been getting in some changes I need for JuliaLowering and trying to clear some of the backlog of things here.

One change which is required for consistency

* #474

Some AST changes which "might be good to do" include

* #466
* #433
* #438
* #488 
* #457

Bugs which are good to fix

* #477
* #316
* #478

Here's some release notes which JuliaRegistrator incorrectly attributed to various 0.4.x releases and will need to be added manually in addition to whatever other release notes:

---

### Merged PRs
- handle ZWJ and emoji sequences, don't break identifiers within graphemes (#372) (@stevengj)
- Generic `sourcefile()` function (#469) (@c42f)
- Move overload of Base.range() to JuliaSyntax.byte_range() (#463) (@c42f)
- Define `numchildren()` function (#464) (@c42f)
- Set version on main branch to 1.0.0-DEV. (#467) (@fredrikekre)
- Allow `Kind`s to be registered by packages outside JuliaSyntax (#461) (@c42f)
- Add the VS Code settings file to .gitignore (#459) (@davidanthoff)
- Hashing fixup, equality support, and serialization support (#452) (@LilithHafner)
- Delete unused down option from parse_docstring (#430) (@LilithHafner)
- Update docs (#400) (@hyrodium)
- Cleanup CI (#454) (@LilithHafner)
- AST: Rearrange `do` to sit inside `call`/`macrocall` (#322) (@c42f)
- Rethinking the `.` AST (#324) (@c42f)

### Closed issues
- Don't define `Base.range(::SyntaxNode)` (#417)
- GreenNode equality and hashing mismatch (#449)
    - `SyntaxNode`s are never equal (#450)
- serialization support (#451)

PR #484, opened 2024-08-07
Clean up and document syntax tree child access API + mark public API

c42f:
Here I commit to a more consistent but simpler child access API for syntax trees, as informed by the JuliaLowering work so far:

* `is_leaf(node)` is given a precise definition (previously `!haschildren()` - but that had issues - see #483)
* `children(node)` returns the child list, or `nothing` if there are no children. The `nothing` might be seen as inconvenient, but mapping across the children of a leaf node is probably an error and one should probably branch on `is_leaf` first.
* `numchildren(node)` is documented
* `node[i]`, `node[i:j]` are documented to index into the child list

We distinguish `GreenNode` and its implementation of `span` from `SyntaxNode` and its implementation of `byte_range` and `sourcetext` - these seem to just have very different APIs, at least as of now.

I've deleted the questionable overloads of multidimensional `getindex` and the `child` function in favor of single dimensional getindex. I don't know whether anyone ever ended up using these. But I didn't and they didn't seem useful+consistent enough to keep the complexity.

I've kept setindex! for now, to set a child of a `SyntaxNode`. Though I'm not sure this is a good idea to support by default and I didn't mark it public yet.

Also I've added more docs!

Also mark the public parts of the API as `public` in Julia 1.11+

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -124,13 +124,46 @@ JuliaSyntax.SHORT_FORM_FUNCTION_FLAG
 
 ## Syntax trees
 
-Syntax tree types:
+Access to the children of a tree node is provided by the functions
+
+```@docs
+JuliaSyntax.is_leaf
+JuliaSyntax.numchildren
+JuliaSyntax.children
+```
+
+For convenient access to the children, we also provide `node[i]`, `node[i:j]`
+and `node[begin:end]` by implementing `Base.getindex()`, `Base.firstindex()` and
+`Base.lastindex()`. We choose to return a view from `node[i:j]` to make it
+non-allocating.
+
+Tree traversal is supported by using these functions along with the predicates
+such as [`kind`](@ref) listed above.
+
+### Trees referencing the source
 
 ```@docs
 JuliaSyntax.SyntaxNode
+```
+
+Functions applicable to `SyntaxNode` include everything in the sections on
+heads/kinds as well as the accessor functions in the source code handling
+section.
+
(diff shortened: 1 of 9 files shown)


timholy (approved):
This looks good. I think TypedSyntax's tests do exploit multidimensional `child`, but that's an easy change. I presume this will become 0.5.0?

c42f:
> I presume this will become 0.5.0?

We need to get to 1.0 so we can use semver properly.  So the next release will be 1.0.0 :-)  I've been working really hard on a lot of 1.0.0 changes in the last few weeks, trying to get the documentation and APIs in better shape.  See also #472, but I keep finding more to do - it turns out there's a lot of changes I want to squeeze in which are informed by the JuliaLowering work.

c42f:
> I think TypedSyntax's tests do exploit multidimensional child, but that's an easy change.

We could keep `child()` if you think it's useful? In that case I'd document what it does and mark it `public`. It's more the multidimensional getindex which is dubious and should probably be deleted.

timholy:
Either way is fine; it's not such a hard change to make to [TypedSyntax's tests](https://github.com/JuliaDebug/Cthulhu.jl/blob/master/TypedSyntax/test/runtests.jl). A regex search suggests there are 74 uses of `child(arg, i, j...)` and that's not too bad to change if you think it would be cleaner without.

c42f:
Ok thanks! Let's go with the more minimal API then :)

Issue #490, opened 2024-08-09
Towards getting JuliaSyntax to work with `--trim` and static binaries.


KristofferC:
Tools that are convenient to use from the command line (like code formatters) are likely to use JuliaSyntax and being able to statically compile these would be nice for latency and distribution. This issue is about the work required to get JuliaSyntax to that state.

Right now, using the branch at https://github.com/JuliaLang/julia/pull/55047 together with the following script:

```julia
tmpdir = mktempdir()
img_path = joinpath(tmpdir, "img.a")
trim_path = joinpath(Sys.BINDIR, "../../test/trimming")

cmd = Base.julia_cmd()
cmd = `$cmd --startup-file=no --history-file=no`
cmd = addenv(`$cmd --output-o $img_path --output-incremental=no --strip-ir --strip-metadata --trim $(joinpath(trim_path,"buildscript.jl")) juliasyntax.jl --output-exe true`, "OPENBLAS_NUM_THREADS" => 1, "JULIA_NUM_THREADS" => 1)

run(cmd)
```

and the file

```julia
module MyApp

using JuliaSyntax

Base.@ccallable function main()::Cint
    JuliaSyntax.parseall(JuliaSyntax.SyntaxNode, "NaN")
    return 0
end

end
```

the following error is quickly encountered:

```
Dynamic call to Base.:(+)(Any, Int64)
In operators.jl:642

Stacktrace:
  [1] +;
    @ operators.jl:642 [inlined]
  [2] seek(io::GenericIOBuffer{GenericMemory{:not_atomic, UInt8, ?}}, n::Int64)
    @ Base iobuffer.jl:288
  [3] parse_julia_literal(txtbuf::Array{UInt8, 1}, head::SyntaxHead, srcrange::UnitRange{Int64})
    @ JuliaSyntax ~/.julia/packages/JuliaSyntax/q0tWf/src/literal_parsing.jl:394
  [4] _to_SyntaxNode(source::SourceFile, txtbuf::Array{UInt8, 1}, offset::Int64, raw::GreenNode{SyntaxHead}, position::Int64, keep_parens::Bool)
    @ JuliaSyntax ~/.julia/packages/JuliaSyntax/q0tWf/src/syntax_tree.jl:72
  [5] #SyntaxNode#106;
    @ ~/.julia/packages/JuliaSyntax/q0tWf/src/syntax_tree.jl:61 [inlined]
  [6] TreeNode;
[...]

PR #491, opened 2024-08-09
add some small type inference improvements to allow for some basic static compiling

KristofferC:
Allows the example in https://github.com/JuliaLang/JuliaSyntax.jl/issues/490 to compile (together with https://github.com/JuliaLang/julia/pull/55047#issuecomment-2277844004).


fredrikekre (approved):
Would be good to add a test case after the base PR is merged.

c42f (approved):
looks good to me.

KristofferC:
> Would be good to add a test case after the base PR is merged.

Yep

Issue #316, opened 2023-06-23
Newlines handled differently compared to old parser


mortenpi:
I am not really sure if this is a bug, or maybe it's a bug in the original parser, but it looks like newlines are handled differently in JuliaSyntax.

This broke Documenter's tests (for some reason it's blacklisted in PkgEval), since we do some exact string checks there for parsed sub-expressions, so the cursor value changes threw those tests off. It's not a problem to handle this with version guards (https://github.com/JuliaDocs/Documenter.jl/pull/2146), but it would be good to know whether this change might change back at some point or not.

Here are some `Meta.parse` calls in 1.9 vs today's master:

```
julia@1.9.1> Meta.parse("x += 3\nγγγ_γγγ\nγγγ\n\n", 22; raise=true)
(:γγγ, 29)

julia@1.10.0-DEV.1556> Meta.parse("x += 3\nγγγ_γγγ\nγγγ\n\n", 22; raise=true)
(:γγγ, 30)
```

```
julia@1.9.1> Meta.parse("x = Int[]\r\n\r\npush!(x, 1)\r\n\n", 14)
(:(push!(x, 1)), 27)

julia@1.10.0-DEV.1556> Meta.parse("x = Int[]\r\n\r\npush!(x, 1)\r\n\n", 14)
(:(push!(x, 1)), 28)
```

```
julia@1.9.1> Meta.parse("x = Int[]\n\npush!(x, 1)\n\n", 12)
(:(push!(x, 1)), 24)

julia@1.10.0-DEV.1556> Meta.parse("x = Int[]\n\npush!(x, 1)\n\n", 12)
(:(push!(x, 1)), 25)
```

```
julia@1.9.1> Meta.parse("1 + 1\n2 + 2\n\n", 7)
(:(2 + 2), 13)

julia@1.10.0-DEV.1556> Meta.parse("1 + 1\n2 + 2\n\n", 7)
(:(2 + 2), 14)
```

```
julia@1.9.1> Meta.parse("1 + 1\n\n\n", 1)
(:(1 + 1), 7)

julia@1.10.0-DEV.1556> Meta.parse("1 + 1\n\n\n", 1)
(:(1 + 1), 9)
```

Slightly more interesting case with a comment line:

```
julia@1.9.1> Meta.parse("1 + 1\n# comment\n\n", 1)
(:(1 + 1), 7)

[...]

PR #493, opened 2024-08-11
Terminate :statement parsing at newlines

timholy:
Addresses a portion of #316

### DIFF
--- a/test/hooks.jl
+++ b/test/hooks.jl
@@ -29,7 +29,7 @@ end
             # Issue #81
             f() = nothing
             """
-        @test JuliaSyntax.core_parser_hook(stmtstr, "somefile", 1, 0, :statement)[2] ∈ 19:21
+        @test JuliaSyntax.core_parser_hook(stmtstr, "somefile", 1, 0, :statement)[2] == 19
     end
 
     @testset "filename and lineno" begin
--- a/test/hooks.jl
+++ b/test/hooks.jl
@@ -11,7 +11,7 @@ function _unwrap_parse_error(core_hook_result)
 end
 
 @testset "Hooks for Core integration" begin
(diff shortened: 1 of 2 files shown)


test/hooks.jl
```diff
@@ -20,6 +20,16 @@ end
...
+
+        # https://github.com/JuliaLang/JuliaSyntax.jl/issues/316#issuecomment-1870294857
+        stmtstr =
+            """
+            plus(a, b) = a + b
+
+            # Issue #81
+            f() = nothing
+            """
+        @test JuliaSyntax.core_parser_hook(stmtstr, "somefile", 1, 0, :statement)[2] ∈ 19:21
```
c42f, line 32: What's the range `19:21` for?

timholy: I don't see that there's a truly unique "right" answer. So it's anything that parses all the code and before the comment starts.

c42f: Ooh I see :)

If we're choosing a particular somewhat-arbitrary but well-defined behavior I'd prefer to be precise with the test so we can detect any unexpected changes.

timholy: Done. I went with 19, which is the earliest this can terminate. I believe this probably matches the flisp parser, but not sure.

I wrote a "wide" test initially because of the annoyances I've experienced over time maintaining tests sensitive to what turn out to be irrelevant details (e.g., whether you add a space between elements when printing tuples). So I've sometimes started to ask, "what behavior do I most care about, and what's the widest test I can write that still enforces that behavior?" But I fully understand where you're coming from here, and it probably makes sense given that this is the *origin* of such choices rather than the *receiver* of such choices.

c42f (approved):
Thanks Tim! LGTM :-)

PR #500, opened 2024-09-03
Remove `Base.convert` methods between `AbstractString` and `Kind`

fredrikekre:


LilithHafner:
This seems worth doing to me. `convert`ing between `Kind` and `String` is too permissive
```julia
julia> push!(["a"], K"=")
2-element Vector{String}:
 "a"
 "="
```

c42f (approved):
This seems great, thanks :)

c42f:
For context, `convert()` was overridden because the implementation of `Kind` was originally inspired by `Base.@enum`.

LilithHafner:
Is squash merge okay?

c42f:
I generally prefer squash merge and I think it's a good default.

Some exceptions where an actual merge commit can be worthwhile:
* The feature branch has several carefully distinguished changes in their own patches such that keeping them separate makes the history easier to understand.
* The change was authored by multiple people in distinct patches
* The merge has nontrivial conflicts with the base branch to the extent that rebasing is very hard and error prone (a very unusual circumstance which mainly happens when multiple large, long lived feature branches interfere. To be avoided! Unlikely to happen in this library.)

fredrikekre:
> For context, convert() was overridden because the implementation of Kind was originally inspired by Base.@enum.

Enums doesn't have any converts like this either though?

PR #501, opened 2024-09-03
Fix serialization of kinds with multibyte chars

fredrikekre:
This patch fixes serialization of `Kind`s to use `sizeof` (number of bytes) instead of `length` (number of characters) when computing number of bytes in the stringified `Kind`.


LilithHafner:
Thanks!

PR #510, opened 2024-09-28
https in readme

spaette:
self descriptive

.svg file not rendering on WiFi networks enforcing https

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # JuliaSyntax
 
 [![Build Status](https://github.com/c42f/JuliaSyntax.jl/workflows/CI/badge.svg)](https://github.com/c42f/JuliaSyntax.jl/actions)
-[![codecov.io](http://codecov.io/github/JuliaLang/JuliaSyntax.jl/coverage.svg?branch=main)](http://codecov.io/github/JuliaLang/JuliaSyntax.jl?branch=main)
+[![codecov.io](https://codecov.io/github/JuliaLang/JuliaSyntax.jl/coverage.svg?branch=main)](https://codecov.io/github/JuliaLang/JuliaSyntax.jl?branch=main)
 
 A Julia compiler frontend, written in Julia.
 


c42f:
thanks :)

spaette:
@c42f 

_off-topic_

don't know if eventually this link will point to somewhere in this repo

```
$ sed -n '213,214p' /tmp/julia/doc/src/base/math.md
The complete list is in the parser code:
<https://github.com/JuliaLang/julia/blob/master/src/julia-parser.scm>
$ 
```

c42f:
> The complete list is in the parser code:
> <https://github.com/JuliaLang/julia/blob/master/src/julia-parser.scm>

Hmm, JuliaSyntax is the default parser so I guess this could be updated. But also it seems we're still in a state of maintaining both parsers at the moment.

spaette:
_on the matter of parsers and lisp_

don't off hand remember who or the location where (JuliaLang repo, and Issue or PR)

but a community member mentioned regarding the transition the subject of s-expressions

the below XS repo includes the source for the then autotools build system version of the code

on my Linux box I had to use the variable `LIBS="-lgccpp"` before running `./configure`

```
$ git clone -q https://github.com/varehus/XS.git
$ sed -n '1377p' XS/configure
  --enable-lisptrees      Allow dumping parsed code in s-expression format
$ sed -n '76,78p' XS/main.cxx
#if LISPTREES
                "       -L      print parser results in LISP format\n"
#endif
$ 
```

at some point the `lisptrees` content was removed from the source code

<hr>

improvements to the source code and converted to the meson build system

a few years ago I built it using a python3 virtual environment (venv) with meson-0.54.x and again with meson-0.56.x

[XS](https://github.com/TieDyedDevil/XS) (meson build system) [ABANDONED] 
[XS-library](https://github.com/TieDyedDevil/XS-library) [ABANDONED]

XS-library/xs/lib.d/math.xs

a Free Software Foundation employee's comments on the foregoing

I looked at math.xs, and boy is it ugly.

In addition, it seems to mean using functional programming to decompose a list.  That is a pain in the neck.

I don't know whether it is more bad or less bad than programming in shell syntax.  But at least there' an excuse in that case: programming is built around a command line syntax.  I see no excuse for making the math.xs syntax so inconvenient.  If they're designing a new language from zero, why take such a big step backard?

PR #511, opened 2024-10-05
Parse docstrings within structs as `K"doc"`

c42f:
Julia's ecosystem (including Base.Docs and flisp lowering) assumes that strings within `struct` definitions are per-field docstrings, but the flisp parser doesn't handle these - they are only recognized when the struct itself has a docstring and are processed by the `@doc` macro recursing into the struct's internals. For example, the following doesn't result in any docs attached to `A`.

```julia
struct A
    "x_docs"
    x

    "y_docs"
    y
end
```

This change adds `K"doc"` node parsing to the insides of a struct, making the semantics clearer in the parser tree and making it possible to address this problems in the future within JuliaLowering.

Also ensure that the `Expr` form is unaffected by this change.

### DIFF
--- a/src/expr.jl
+++ b/src/expr.jl
@@ -466,6 +466,19 @@ function _internal_node_to_Expr(source, srcrange, head, childranges, childheads,
         headsym = :call
         pushfirst!(args, :*)
     elseif k == K"struct"
+        @assert args[2].head == :block
+        orig_fields = args[2].args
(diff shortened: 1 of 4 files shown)


c42f:
I think this should be pretty safe and reasonable (comments welcome though)

It's technically breaking for `SyntaxNode` but I still haven't released 1.0 soo :sweat_smile: ahh

PR #513, opened 2024-10-14
support for for-else and while-else

simeonschaub:
companion to JuliaLang/julia#56153


test/expr.jl
```diff
@@ -189,6 +189,21 @@
...
+                              LineNumberNode(2),
+                              :y,
+                         ),
+                         Expr(:block,
+                              LineNumberNode(4),
+                              :z,
+                              LineNumberNode(5)
+                         )
+                    )
+            end
```
c42f, line 206:
```suggestion

            @test parsestmt("for x=xs\ny\nelse\nz\nend", version=v"1.12") ==
                Expr(:for,
                     Expr(:(=), :x, :xs),
                     Expr(:block,
                          LineNumberNode(2),
                          :y,
                     ),
                     Expr(:block,
                         LineNumberNode(4),
                          :z,
                          LineNumberNode(5)
                     )
                )

```

c42f (commented):
The upstream reception of this seems overall negative, so I guess we might not do it?

A minor comment if we do ever merge this - for tests we should never depend on `VERSION` unless really necessary: JuliaSyntax is multi-version aware :)

c42f:
Heya, we just moved JuliaSyntax into Base in https://github.com/JuliaLang/julia/pull/59870 - if you want to re-propose this PR over there, I've created a branch https://github.com/JuliaLang/JuliaSyntax.jl/tree/pr-for-Base/513 to make this easier.

* Clone JuliaLang/julia
* Check out pr-for-Base/513
* Rebase on top of master

For example:

```bash
# cd julia_dir

git fetch JuliaSyntax
git checkout pr-for-Base/513
git rebase origin/master
```

PR #518, opened 2024-12-06
Fix `incomplete_tag` generation with trailing whitespace

### LINKED ISSUE #512: Regression in incomplete_tag
On JuliaSyntax master:
```
julia> s = "CompletionFoo.test(1, 1, "
"CompletionFoo.test(1, 1, "

julia> Base.incomplete_tag(Meta.parse(s, raise=false, depwarn=false))
:none
```
vs Base:
```
julia> Base.incomplete_tag(Meta.parse(s, raise=false, depwarn=false))
:other
```

I looked into adjusting the tag selector a bit, but couldn't come up with a good version that also distinguished `[x ` (with space), `[x)` and `(` and `( `, so there may need to be some adjustment to the source-location representation itself. @c42f can you take a look?

c42f:
Previously we relied on the last SyntaxNode consuming all trailing whitespace when detecting incomplete syntax. However this assumption was broken by #397 and is generally fragile with respect to any extra bump_trivia() calls. Fix this by just comparing to the stream position before any bumping of remaining trivia.

Should fix #512

### DIFF
--- a/src/hooks.jl
+++ b/src/hooks.jl
@@ -170,6 +170,7 @@ function core_parser_hook(code, filename::String, lineno::Int, offset::Int, opti
             end
         end
         parse!(stream; rule=options)
+        pos_before_trivia = last_byte(stream)
         if options === :statement
             bump_trivia(stream; skip_newlines=false)
             if peek(stream) == K"NewlineWs"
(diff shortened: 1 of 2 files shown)


Keno:
The additional test cases I had were "CompletionFoo.test(1, 1, " (which is the one that base depends on so good to CI test here) and "[x)". They both look good here so LGTM.

c42f:
Ah now I see.

The `CompletionFoo` stuff is all in the REPL stdlib tests and tests a lot more than the parser.

But the JuliaSyntax tests are drawn from Base tests in test/client.jl (plus additional cases I constructed).

PR #522, opened 2024-12-22
Make left hand side of `->` parse as a tuple of arguments

c42f:
This makes argument lists on the left hand side of `->` parse as tuples. In particular, this fixes a very inconsistent case where the left hand side previously parsed as a `K"block"`; we now have:

* `(a;x=1)->b` ==> `(-> (tuple-p a (parameters (= x 1))) b`

rather than `(block a (= x 1))` on the left hand side.

In addition, the following forms are treated consistently

* `a->b`       ==> `(-> (tuple a) b)`
* `(a)->b`     ==> `(-> (tuple-p a) b)`
* `(a,)->b`    ==> `(-> (tuple-p-, a) b)`

The upside of this is that expression processing of `->` syntax should be much easier.

There's one aberrant case involving `where` which is particularly difficult and still not dealt with:

`(x where T) -> b` does not parse as `(where (tuple x) T)` on the left hand side. However, `where` precedence involving `::` and `->` is already horribly broken and this syntax will always be awful to write unless we make breaking changes. So I'm too tempted to call this a lost cause for now 😬.

Compatibility shims for converting the `SyntaxNode` form back to `Expr` in order to keep `Expr` stable are included. (At some point we should consider fixing this and deleting these shims because the new form is so much more consistent and would be reflected neatly into `Expr` form.)

### DIFF
--- a/docs/src/reference.md
+++ b/docs/src/reference.md
@@ -70,6 +70,7 @@ class of tokenization errors and lets the parser deal with them.
 * Standalone dotted operators are always parsed as `(. op)`. For example `.*(x,y)` is parsed as `(call (. *) x y)` (#240)
 * The `K"="` kind is used for keyword syntax rather than `kw`, to avoid various inconsistencies and ambiguities (#103)
 * Unadorned postfix adjoint is parsed as `call` rather than as a syntactic operator for consistency with suffixed versions like `x'ᵀ` (#124)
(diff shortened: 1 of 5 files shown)


KristofferC:
This caused an error in RBNF.jl, see https://github.com/thautwarm/RBNF.jl/pull/17 for a workaround. I guess this is https://github.com/JuliaLang/JuliaSyntax.jl/issues/535.

Issue #368, opened 2023-10-15
recursion causes stackoverflow and crash


vtjnash:
observed in https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_hash/5a9df24_vs_3250804/ODE.primary.log

In general, a compiler/parser commonly must not use recursion since it is surprisingly easy for the user to generate inputs that cause it to crash.

```
[...]

PR #524, opened 2025-01-03
WIP: Non-recursive shunting yard algorithm for expression parsing

c42f:
This is very much a work in progress, but shows some promise for *very greatly* reducing our recursion depth. The idea is to use the non recursive shunting-yard algorithm for parsing operators and grouping-parentheses but to delegate back to the existing recursive formulation for other constructs.

This will likely solve #368 in all practical cases - I expect deeply recursive constructs only for huge chains of operators and parentheses.

Currently our operator parsing consumes maybe 15 or so stack frames every time a grouping parenthesis nested in combination with arithmetic. This quickly leads to absurdly deep program stacks and stack overflow. Moving to a system like a Pratt parser where we skip non-used precedence levels would make this a single stack frame. Moving to the shunting yard algorithm makes it zero stack frames, provided we can also use it to treat grouping parentheses (not an entirely simple thing, because parentheses in Julia are *very* syntactically overloaded.)

The biggest challenge here is to ensure we exactly reproduce all of Julia's operator precedence rules, which have many complicated special cases. The demo here doesn't cover many special cases, but it does show how a few of these can be dealt with quite simply in the non-recursive context. For example, chains of `+` and `*` need to parse into a single n-ary call, and it was reasonably easy to add this special case.

### DIFF
--- a/src/parser.jl
+++ b/src/parser.jl
@@ -589,6 +589,186 @@ function parse_eq_star(ps::ParseState)
     end
 end
 
+function binding_power(tok) # TODO: prefix ops
+    k = kind(tok)                      # lhs rhs
+    return k in KSet"NewlineWs EndMarker" ? (0,   0) :
+           k == K"("                 ? (1000, 1)     :
+           k == K")"                 ? (1,    1001)  :
+           k > K"BEGIN_OPS" ? (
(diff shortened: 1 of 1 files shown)


c42f:
If we want to move this to Base sometime ...

```bash
# cd julia_dir

git fetch JuliaSyntax
git checkout pr-for-Base/524
git rebase origin/master
```

PR #525, opened 2025-01-08
Add U+1F8B2 🢲 as an operator

pthariensflame:
The character U+1F8B2 🢲 (RIGHTWARDS ARROW WITH LOWER HOOK) is new is Unicode 16; it is of a kind (no pun intended) with the longstanding characters U+21A9 ↩ (LEFTWARDS ARROW WITH HOOK) and U+21AA ↪ (RIGHTWARDS ARROW WITH HOOK), both of which are already supported as operators in Julia.  If more evidence of its worth is needed, it was added to Unicode as part of the Symbols for Legacy Computing effort, wherein it was sourced from Smalltalk character sets in the 1970s—so it has a very long history of being used in programming languages!

### DIFF
--- a/src/kinds.jl
+++ b/src/kinds.jl
@@ -479,6 +479,7 @@ register_kinds!(JuliaSyntax, 0, [
         "↶"
         "↺"
         "↻"
+        "🢲"
     "END_ARROW"
 
     # Level 4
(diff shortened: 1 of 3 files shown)


c42f:
This seems fine to add, thanks.

However, this patch doesn't work as-is, due to an unfortunate combination of circumstances.  In particular, we still rely on `Base.is_id_start_char('🢲')` which unfortunately returns `true` for this arrow.

It's easiest just to special-case this arrow for now so I'll do that (we should really do #381, though :grimacing: )

PR #537, opened 2025-02-11
Parse arr[end] differently from arr[var"end"]

mlechu:
Fixes a small issue present in both parsers (see discussion at https://github.com/JuliaLang/julia/issues/57269) due to treating `begin`/`end` in indexing expressions as normal identifiers. 

This would be a breaking change that goes with PR https://github.com/JuliaLang/julia/pull/57368, which makes the same change to the lisp parser, and changes lowering to expect `Expr(:end)` instead of a symbol.

### DIFF
--- a/src/parser.jl
+++ b/src/parser.jl
@@ -3627,14 +3627,10 @@ function parse_atom(ps::ParseState, check_identifiers=true, has_unary_prefix=fal
         elseif check_identifiers && is_closing_token(ps, leading_kind)
             # :(end)  ==>  (quote-: (error end))
             bump(ps, error="invalid identifier")
-        elseif ps.end_symbol && leading_kind in KSet"begin end"
+        elseif ps.end_symbol && leading_kind in KSet"begin end" && ps.stream.version >= (1, 12)
             # https://github.com/JuliaLang/julia/issues/57269
             # Parse a[begin] differently from a[var"begin"]
-            if leading_kind == K"begin"
-                bump(ps, remap_kind=K"RefBegin")
-            elseif leading_kind == K"end"
-                bump(ps, remap_kind=K"RefEnd")
-            end
+            bump(ps)
         else
             # Remap keywords to identifiers.
             # :end  ==>  (quote-: end)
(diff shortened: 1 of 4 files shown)


src/parser.jl
```diff
@@ -3627,6 +3627,14 @@ function parse_atom(ps::ParseState, check_identifiers=true, has_unary_prefix=fal
         elseif check_identifiers && is_closing_token(ps, leading_kind)
             # :(end)  ==>  (quote-: (error end))
             bump(ps, error="invalid identifier")
+        elseif ps.end_symbol && leading_kind in KSet"begin end"
```
Keno, line 3630: Needs a version check on the parse syntax, since JuliaSyntax itself is not tied to a Julia version.

c42f:
I think the changes to `SyntaxNode` are good here and we should do them even if it's too breaking to change `Expr` in Base.  JuliaSyntax already contains many such AST differences - some of which are much more significant than changing begin/end. (As an example, consider `K"dotcall"`.)

So if you like we could merge this right away if you make sure the `Expr` form remains unchanged. We just need:

* Tests added to `parsestmt_with_kind_tests` in test/parser.jl to assert the correct kind for these tokens
* Currently-failing tests fixed
* Reuse `K"begin"` / `K"end"` rather than make new kinds. Keeping close to the source strings makes it super easy to remember the right kind.

Later you can make a separate PR if PkgEval deems the `Expr` changes not to be breaking.

src/parser.jl
mlechu: Thanks! Assuming 1.12?

mlechu:
Thanks for the help! 


> * Tests added to `parsestmt_with_kind_tests` in test/parser.jl to assert the correct kind for the begin/end leaf nodes in the tree

Am I correct in thinking I would need to update `parsestmt_with_kind_tests` to take the version into account?

src/parser.jl
Keno: Let's do 1.13 - we can change it later if we decide to backport.

c42f:
Heya, I've created a branch https://github.com/JuliaLang/JuliaSyntax.jl/tree/pr-for-Base/537 to make this easier to move into Base

* Clone JuliaLang/julia
* Check out pr-for-Base/537
* Rebase on top of master

For example:

```bash
# cd julia_dir

git fetch JuliaSyntax
git checkout pr-for-Base/537
git rebase origin/master
```

PR #538, opened 2025-02-13
Make `Expr(:incomplete)` detection more robust to whitespace

### LINKED ISSUE #519: ParseError when typing multi-line input with comment
Since the backport of #397 to `release-0.4` (julia nightly since https://github.com/JuliaLang/julia/pull/56110) and also on `main`, I cannot enter multi-line input which contains comments. Once I press return after the comment it directly aborts with a `ParseError` instead of letting me continue the input in the next line:
```julia
julia> function bla()
       a=1
       b=2 #a comment
ERROR: ParseError:
# Error @ REPL[2]:3:4
a=1
b=2 #a comment
#  └ ── Expected `end`
Stacktrace:
 [1] top-level scope
   @ REPL:1
```
The same happens when entering an array in multiple lines:
```julia
julia> arr = [
       1,
       2,
       3, # i like this number
ERROR: ParseError:
# Error @ REPL[3]:4:3
2,
3, # i like this number
# └ ── Expected `]` or `,`
Stacktrace:
 [1] top-level scope
   @ REPL:1
```
Note that this applies only to entering this manually line by line, pasting multi-line strings with comments works fine.

With previous julia nightly (or JuliaSyntax.jl) versions the error appears only on the second consecutive empty line:
```julia
julia> arr = [
       1,
       2,
       3, # i like this number
       
       
ERROR: ParseError:
# Error @ REPL[33]:5:2
3, # i like this number

#└ ── Expected `]`
Stacktrace:
 [1] top-level scope
   @ REPL:1
```

c42f:
Rework incomplete expression detection so that trailing whitespace is always ignored, regardless of how the parser itself decides to attach it to other nodes of the tree.

To do this, we walk back from the end of the parse stream and look for the byte offset of the last non-whitespace token. We then use that to determine whether the error node is "at the end of the parse".

Improve testing by
* Extracting the incomplete expressions which are part of the REPL stdlib tests and ensuring these match the incomplete tag generation of the flisp parser. Fix some divergences for `var""` syntax and invalid escape sequences in strings.
* Ensuring that we test both `:statement` and `:all` level parsing - the REPL uses `:all` to allow parsing of multiple top level statements, so we need to test this.

Also fix a minor bug where `enable_in_core!(false)` would result in the flisp parser being used, regardless of whether `VERSION` ships with JuliaSyntax enabled by default.

Fixes #519. See also #518.

### DIFF
--- a/src/hooks.jl
+++ b/src/hooks.jl
@@ -35,15 +35,17 @@ end
 function _incomplete_tag(n::SyntaxNode, codelen)
     i,c = _first_error(n)
     if isnothing(c) || last_byte(c) < codelen || codelen == 0
-        return :none
-    elseif first_byte(c) <= codelen
-        if kind(c) == K"ErrorEofMultiComment" && last_byte(c) == codelen
+        if kind(c) == K"ErrorEofMultiComment"
             # This is the one weird case where the token itself is an
             # incomplete error
             return :comment
         else
             return :none
         end
+    elseif first_byte(c) <= codelen && kind(c) != K"ErrorInvalidEscapeSequence"
+        # "ErrorInvalidEscapeSequence" may be incomplete, so we don't include it
+        # here as a hard error.
+        return :none
     end
     if kind(c) == K"error" && numchildren(c) > 0
         for cc in children(c)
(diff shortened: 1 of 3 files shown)


fingolfin:
Thank you!

Is there a chance this will be in Julia 1.12?

c42f:
Yes, this is a bug fix so it should go into 1.12.

PR #547, opened 2025-04-09
add `activate!` function as alias to `enable_in_core!`

oscardssmith:
This will allow for automatic hookup to be done via `InteractiveUtils.@activate JuliaSyntax`. 

TODO: On julia's end, we'll need to add JuliaSyntax to the components we're allowed to activate (and bump the JuliaSyntax version).

### DIFF
--- a/src/expr.jl
+++ b/src/expr.jl
@@ -580,8 +580,8 @@ end
 
 function build_tree(::Type{Tuple{Expr, SyntaxNode}}, stream::ParseStream;
                     filename=nothing, first_line=1, kws...)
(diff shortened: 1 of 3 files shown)


oscardssmith:
@mlechu now that the lowering parts of this are removed, is it ready for review/merge?

@c42f Any thoughts on this approach?

mlechu:
> ready for review/merge?

Yup, should be good for review.  I think the activate macro was modelled after our `enable_in_core!` anyway.

aviatesk (approved):

PR #552, opened 2025-05-13
Make `source_location()` and SourceFile's and `line_starts` consistent when the last source byte is a newline

mlechu:
When constructing SourceFile's `line_starts`, we currently pretend there's a
newline at the end of the source if there isn't one, and then ignore the last
newline unconditionally.  This became a problem in LSP testing, where calling
`source_location` with a cursor position at the end of the file would give a
wrong result only if the last character was a newline.

```
julia> source_location(SourceFile("\n\n\n"), 1)
(1, 1)

julia> source_location(SourceFile("\n\n\n"), 2)
(2, 1)

julia> source_location(SourceFile("\n\n\n"), 3)
(3, 1)

julia> source_location(SourceFile("\n\n\n"), 4)
(3, 2)
```

This change removes the conditional line and the code stripping it.  This also
brings `line_starts` closer to its definition (one entry per line in the file).

Alternatively, we could add the dummy line unconditionally.  It would make for
calculating the length of the last real line easier, but would make
`line_starts` less explainable.

### DIFF
--- a/src/source_files.jl
+++ b/src/source_files.jl
@@ -156,9 +156,6 @@ function SourceFile(code::AbstractString; filename=nothing, first_line=1,
         # The line is considered to start after the `\n`
         code[i] == '\n' && push!(line_starts, i+1)
     end
-    if isempty(code) || last(code) != '\n'
-        push!(line_starts, ncodeunits(code)+1)
-    end
     SourceFile(code, first_index-1, filename, first_line, line_starts)
 end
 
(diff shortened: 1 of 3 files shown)


aviatesk (approved):
This behavior feels more natural for what `source_location` returns, and it also seems like the implementation gets a bit simpler as a result.

aviatesk:
@c42f I'm hoping to move this PR forward and release a new patch version soon. Could you give us some comments on this if you have any?

PR #553, opened 2025-05-13
simple_hash faster by lifting length(str)

adienes:
avoids github.com/JuliaLang/julia/issues/33712, matches https://github.com/JuliaLang/Tokenize.jl/pull/213

probably rarely makes a difference but might as well; this should be strictly better.

### DIFF
--- a/src/tokenize.jl
+++ b/src/tokenize.jl
@@ -1339,7 +1339,8 @@ end
 function simple_hash(str)
     ind = 1
     h = UInt64(0)
-    while ind <= length(str)
+    L = length(str)
+    while ind <= L
         h = simple_hash(str[ind], h)
         ind = nextind(str, ind)
     end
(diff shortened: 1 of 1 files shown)


simeonschaub:
I don't think using `length` here is even correct for non-ASCII strings, this code should be using `lastindex` instead

src/tokenize.jl
```diff
@@ -1339,7 +1339,8 @@ end
 function simple_hash(str)
     ind = 1
     h = UInt64(0)
-    while ind <= length(str)
+    L = length(str)
+    while ind <= L
```
pfitzseb, line 1343:
```suggestion
    L = min(lastindex(str), MAX_KW_LENGTH)
    while ind <= L
```
is probably more correct and cheaper.

pfitzseb (approved):

PR #561, opened 2025-06-09
kinds: rm unused kind K"inert"

Keno:
`inert` is mostly used in lowering. As far as I can tell, there is precisely one place in the flisp parser that produces it:

https://github.com/JuliaLang/julia/blob/d6b3669621ceb18aea693d8544b2c38870d289ad/src/julia-parser.scm#L1259

However, we do not insert and `inert` there in JuliaSyntax, instead preferring to automatically quote the rhs of a `.` head in expr conversion:

https://github.com/JuliaLang/JuliaSyntax.jl/blob/eceaa39706c6ecc3ef496613096216625dba558e/src/expr.jl#L310

As a result, (and as code coverage complains),  the syntax head is unused and should be removed here. However, it should probably then be added back in JuliaLowering.

### DIFF
--- a/src/expr.jl
+++ b/src/expr.jl
@@ -526,8 +526,6 @@ end
     elseif k == K"module"
         pushfirst!(args, !has_flags(nodehead, BARE_MODULE_FLAG))
         pushfirst!((args[3]::Expr).args, loc)
-    elseif k == K"inert"
-        return QuoteNode(only(args))
     elseif k == K"quote"
         if length(args) == 1
             a1 = only(args)
--- a/src/kinds.jl
+++ b/src/kinds.jl
@@ -1015,7 +1015,6 @@ register_kinds!(JuliaSyntax, 0, [
         "dotcall"
         "comparison"
         "curly"
-        "inert"          # QuoteNode; not quasiquote
         "juxtapose"      # Numeric juxtaposition like 2x
         "string"         # A string interior node (possibly containing interpolations)
         "cmdstring"      # A cmd string node (containing delimiters plus string)


Keno:
👍 I'm assuming JuliaLowering must understand this and for flisp lowering, `QuoteNode`, gets turned into `(inert)`, so what expr.jl is doing here will look the same to flisp lowering.

mlechu:
Looks fine to me, though I hope for a day these can be in Base

Keno:
> though I hope for a day these can be in Base

Isn't `QuoteNode` in the AST exactly this?

Keno:
Or were you saying you want syntax for it?

mlechu:
We might want to revert this.  I didn't know at the time that JuliaLowering uses expr.jl for (pre-lowering) SyntaxTree->Expr conversion.  

K"inert" won't be in the tree if we got it from normal parsing, but if it's been through macro expansion (or is from expr->syntaxtree, depending on whether we end up back-converting QuoteNodes) it can show up.

mlechu:
Maybe.  `JuliaSyntax.node_to_expr` is only defined on pre-lowering trees, so JuliaLowering shouldn't be adding new forms.  I was thinking that all pre-lowering (post-parsing up to post-macroexpand) trees should allow the same set of forms, so `node_to_expr` being defined in one piece is best.

On the other hand, I'll probably be setting up a JuliaLowering layer to `node_to_expr` anyway for macro hygiene (i.e. do a subset of the `gynsym`ing we do normally) since that doesn't make sense to put in the parser.  We could also deal with K"inert" there, but other than requiring gensyms, I don't think the post-parse and post-macroexpand tree definitions will (or should) drift too far apart.

cc @c42f in case you have opinions

Issue #536, opened 2025-02-06
Separating the basic parsing infrastructure from the concrete Julia parser (or BYOT, Bring Your Own Tokenizer)


KristofferC:
There is a lot of nice infrastructure to write custom parsers in JuliaSyntax.jl and there is an example of that in e.g. https://github.com/JuliaLang/JuliaSyntax.jl/blob/main/prototypes/simple_parser.jl. The issue is that the tokenizer is hardcoded

https://github.com/JuliaLang/JuliaSyntax.jl/blob/86bc4331eaa08e08bf2af1ba7b50bbbf4af70cdb/src/parse_stream.jl#L345-L346

and there is a bit of a mix of generic parsing functionality and julia code specific things like

https://github.com/JuliaLang/JuliaSyntax.jl/blob/86bc4331eaa08e08bf2af1ba7b50bbbf4af70cdb/src/parse_stream.jl#L115

An idea could be to try extract the language agnostic parts of the parser into a separate module/package and make the julia parser an implementation on top of this. Someone who wants to use the infrastructure for a different language could then write their own lexer but still have use for all the other parsing utilities in here. 

This would for example require defining an interface for what a custom lexer (and token) should support and e.g. replace hard-coded checks like

https://github.com/JuliaLang/JuliaSyntax.jl/blob/86bc4331eaa08e08bf2af1ba7b50bbbf4af70cdb/src/parse_stream.jl#L1264-L1266

with generic versions of this like `thombstron(TokenType)` instead of `K"THOMBSTONE" etc.

With the exception of the tokenizer I am not familiar with the code base so I don't know the level of effort and how feasible this is but I thought I would float the idea. The use-case we have is to use JuliaSyntax.jl to parse another language and still have access to e.g. good source code location etc which the data structures in here were designed to provide.

c42f:
Here's the parts which are reusable, in principle:
* `ParseStream` and its associated types and generic functions `emit() / bump()` in `parse_stream.jl` - the majority of this is general and some of it is quite subtle to implement (eg, `build_tree()`).
* SourceFile and pretty printing of error messages
* Data structures for diagnostics
* The infrastructure for creating and working with `Kind` (but not `Kind` itself - that's Julia-specific)
* Potentially `GreenNode`/`SyntaxNode` if anyone wants to use those tree types
* Helper functions for converting strings to values (literal_parsing.jl)
* Probably some infrastructure within the tokenizer, but not the tokenizer itself

[...]

PR #563, opened 2025-06-09
RFC/WIP: Reorganize into largely self-container pieces

Keno:
A I mentioned in #560, and as contemplated in #536, I'd like to try re-using
JuliaParser infrastructure to replace parsers I've written for some other languages.
This takes the first step to do so by moving various files into directories depending
on whether they are language-dependent or not. Right now there is still some coupling
and of course, there are no actual abstractions between these pieces. The idea would
be to intrduce those over time. For now, if we put in this refactoring, the way to
use this would be to copy the appropriate pieces (at least `core/`) into your
downstream parser and then rewrite it to those APIs. I'm planning to do that with
a parser or two to see if I hit any big API issues and see what it would take
to actually make the re-use happen.


KristofferC:
I think this is a good start. The ultimate goal should probably be a total split the package into something like Syntax.jl + JuliaSyntax.jl where Syntax.jl has the core stuff and JuliaSyntax.jl implements the julia tokenizer + parser on top of that.

Issue #567, opened 2025-06-28
[META] Planned changes to AST structure


Keno:
Discussing with @mlechu, I think there's a few near term changes we'd like to make to the core JuliaSyntax AST data structures. I'm opening up this issue so we have a place to track them all and discuss the overall plan. Particulars of any change can go to the appropriate issue or PR. I think the plan would be that we put them all in and then tag JuliaSyntax 2.0 for 1.13 to go with putting JuliaLowering in Base.

Planned changes (will amend list as more things come up).

1. [ ]  #564
2. [ ] Remove the DOTOP_FLAG - it's hardly used at this point anyway
3. [ ] Find a different representation for the numeric bits - they are only used by two heads, but take up 8 bits in every syntax node. This seems a bit wasteful.

Keno:
> 1. Find a different representation for the numeric bits - they are only used by two heads, but take up 8 bits in every syntax node. This seems a bit wasteful.

I did this change locally, but on further thought, I think we can and should just overlap these with the other head-specific flags in the top 8 bits of the flags. I don't see any other compelling reason to move these out.

mlechu:
Overlap is fine, but I still think it would be bothersome to get or set the numeric flags in macros accepting or producing those forms

Keno:
Is the same not true for the other head-specific flags? Do we have concrete examples of what macros will look like in the new system? I think that could help guide API design.

mlechu:
Fair, it makes sense to wait until that's sorted out.  Some examples of the current state of new macros are [here](https://github.com/c42f/JuliaLowering.jl/blob/main/test/macros.jl)

PR #568, opened 2025-07-01
Remove DOTOP_FLAG from tokenizer and parser

Keno:
This implements the first of a series of AST format changes described in #567. In particular, this removes the DOTOP_FLAG. Currently, we already do not emit the DOTOP_FLAG on terminals, they always get split into one dot token and one identifier token (although we did set it on the intermediate tokens that came out of the lexer). The only four kinds for which DOTOP_FLAG was ever set in the final AST were `=`, `op=`, `&&` and `||`. This introduces separate head kinds for each of these (similar to how there are already separate head calls for `dotcall` and `dot). Otherwise the AST structure should be unchanged.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,17 +1,17 @@
 The JuliaSyntax.jl package is licensed under the MIT "Expat" License:
 
 > Copyright (c) 2021 Julia Computing and contributors
-> 
+>
 > Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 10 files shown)


Keno:
> Then we only have import left, which is basically its own thing anyway and could possibly be its own lex context.

This doesn't work because import can take interpolations, which can have arbitrary syntax in them, so you can't lex it different.

Issue #334, opened 2023-07-28
We don't need 800 kinds 


LilithHafner:
"even Claire said so" –@c424

These don't all need different kinds:
```
...
        "⥺"
        "⭇"
        "⭈"
        "⭉"
        "⭊"
        "⭋"
        "⭌"
        "￩"
        "￫"
        "⇜"
        "⇝"
        "↜"
        "↝"
        "↩"
        "↪"
...
```

ShalokShalom:
So.. remove some of them? 
All?
Whats the purpose of this issue?

LilithHafner:
Seems reasonable to have one kind per precedence level, or even one kind for all standard infix operators.
I filed this issue to log some technical debt/code smell. It would be nice to clean this up.

c42f:
Having mentioned this elsewhere I thought I'd expand just a little on the design we were discussing when this was originally written.

Distinct `Kind`s are only really useful to the extent that they impact parsing.  The only way that operators generally impact parsing is through their precedence classes. All except for a few specific operator kinds which are treated specially during parsing, such as `K"+"` for `parse_with_chains()`.

So we can get away with very few kinds for operators - at max we need one `Kind` per precedence class plus one kind per special operator type.

If we want to go even further, it may even be possible to reduce all operators to `K"Identifier"` and emit the precedence class into the numeric flags. Then there's no more `remap_kind` for operators which could neaten a few things up in the parser.

PR #573, opened 2025-07-10
Stop emitting K".." and K"..." in lexer

Keno:
Unfortunately, the sequences `..` and `...` do not always refer to the `..` operator or the `...` syntax. There are two and a half cases where they don't:

1. After `@` in macrocall, where they are both regular identifiers
2. In `import ...A` where the dots specify the level
3. `:(...)` treats `...` as quoted identifier

Case 1 was handled in a previous commit by lexing these as identifiers after `@`.

However, as a result of case 2, it is problematic to tokenize these dots together; we essentially have to untokenize them in the import parser. It is also infeasible to change the lexer to have speical context-sensitive lexing in `import`, because there could be arbitrary interpolations, `@eval import A, $(f(x..y)), ..b`, so deciding whether a particular `..` after import refers to the operator or a level specifier requires the parser.

Currently the parser handles this by splitting the obtained tokens again in the import parser, but this is undesirable, because it invalidates the invariant that the tokens produced by the lexer correspond to the terminals of the final parse tree.

This PR attempts to address this by only ever having the lexer emit `K"."` and having the parser decide which case it refers to. The new non-terminal `K"dots"` handles the identifier cases (ordinary `..` and quoted `:(...)` ). K"..." is now exclusively used for splat/slurp, and is no longer used in its non-terminal form for case 3.

### DIFF
--- a/src/integration/expr.jl
+++ b/src/integration/expr.jl
@@ -338,6 +338,9 @@ end
         return adjust_macro_name!(retexpr.args[1], k)
     elseif k == K"?"
         retexpr.head = :if
+    elseif k == K"DotsIdentifier"
+        n = numeric_flags(flags(nodehead))
+        return n == 2 ? :(..) : :(...)
     elseif k == K"op=" && length(args) == 3
         lhs = args[1]
         op = args[2]
--- a/src/julia/julia_parse_stream.jl
+++ b/src/julia/julia_parse_stream.jl
@@ -137,8 +137,8 @@ function untokenize(head::SyntaxHead; unique=true, include_flag_suff=true)
         is_postfix_op_call(head) && (str = str*"-post")
 
         k = kind(head)
-        # Handle numeric flags for nrow/ncat nodes
-        if k in KSet"nrow ncat typed_ncat"
+        # Handle numeric flags for nodes that take them
+        if k in KSet"nrow ncat typed_ncat DotsIdentifier"
             n = numeric_flags(head)
             n != 0 && (str = str*"-"*string(n))
         else
@@ -301,13 +301,18 @@ function bump_split(stream::ParseStream, split_spec::Vararg{Any, N}) where {N}
     return position(stream)
 end
 
-function peek_dotted_op_token(ps, allow_whitespace=false)
+function peek_dotted_op_token(ps)
     # Peek the next token, but if it is a dot, peek the next one as well
     t = peek_token(ps)
     isdotted = kind(t) == K"."
     if isdotted
         t2 = peek_token(ps, 2)
-        if !is_operator(t2) || (!allow_whitespace && preceding_whitespace(t2))
+        if preceding_whitespace(t2)
+            isdotted = false
+        elseif !is_operator(t2)
+            isdotted = false
+        elseif kind(t2) == K"." && peek(ps, 3) == K"."
+            # Treat `..` as dotted K".", unless there's another dot after
             isdotted = false
         else
             t = t2
@@ -316,13 +321,21 @@ function peek_dotted_op_token(ps, allow_whitespace=false)
     return (isdotted, t)
 end
 
-function bump_dotted(ps, isdot, flags=EMPTY_FLAGS; emit_dot_node=false, remap_kind=K"None")
+function bump_dotted(ps, isdot, t, flags=EMPTY_FLAGS; emit_dot_node=false, remap_kind=K"None")
     if isdot
-        if emit_dot_node
-            dotmark = position(ps)
-            bump(ps, TRIVIA_FLAG) # TODO: NOTATION_FLAG
-        else
-            bump(ps, TRIVIA_FLAG) # TODO: NOTATION_FLAG
+        dotmark = position(ps)
+        bump(ps, TRIVIA_FLAG)
+        if kind(t) == K"."
+            # .. => DotsIdentifier-2
+            bump(ps, TRIVIA_FLAG)
+            pos = emit(ps, dotmark, K"DotsIdentifier", set_numeric_flags(2))
+            nt = peek_token(ps)
+            if is_operator(nt) && !preceding_whitespace(nt)
+                # a..+b  => (call-i a .. (error-t) (call + b))
+                bump_invisible(ps, K"error", TRIVIA_FLAG,
+                    error="`..` here is interpreted as a binary operator. A space is required if followed by another operator.")
+            end
+            return pos
         end
     end
     pos = bump(ps, flags, remap_kind=remap_kind)
--- a/src/julia/kinds.jl
+++ b/src/julia/kinds.jl
@@ -278,8 +278,6 @@ register_kinds!(JuliaSyntax, 0, [
     "ErrorInvalidOperator"
     "Error**"
 
-    "..."
-
     # Level 1
     "BEGIN_ASSIGNMENTS"
         "BEGIN_SYNTACTIC_ASSIGNMENTS"
@@ -774,7 +772,6 @@ register_kinds!(JuliaSyntax, 0, [
     # Level 8
     "BEGIN_COLON"
         ":"
-        ".."
         "…"
         "⁝"
         "⋮"
@@ -1033,6 +1030,10 @@ register_kinds!(JuliaSyntax, 0, [
         "typed_ncat"
         "row"
         "nrow"
+        # splat/slurp
+        "..."
+        # ../... as a identifier
+        "DotsIdentifier"
         # Comprehensions
         "generator"
         "filter"
--- a/src/julia/parser.jl
+++ b/src/julia/parser.jl
@@ -371,7 +371,7 @@ function parse_RtoL(ps::ParseState, down, is_op, self)
     down(ps)
     isdot, tk = peek_dotted_op_token(ps)
     if is_op(tk)
-        bump_dotted(ps, isdot, remap_kind=K"Identifier")
+        bump_dotted(ps, isdot, tk, remap_kind=K"Identifier")
         self(ps)
         emit(ps, mark, isdot ? K"dotcall" : K"call", INFIX_FLAG)
     end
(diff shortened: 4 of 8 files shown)


c42f:
> Currently the parser handles this by splitting the obtained tokens again in the import parser, but this is undesirable, because it invalidates the invariant that the tokens produced by the lexer correspond to the non-terminals of the final parse tree.

Assuming you mean terminals of the parse tree here...

So - I do agree that `bump_split` is super ugly and it's great to delete it.  On the other hand, I have much less of an issue with `bump_glue()`, because pasting tokens together doesn't require peeking at the text inside them.

IMO, this is actually the main invariant we should aim for when designing token kinds: the parser should never need to look at the text inside a token in order to produce the parse tree. And this is the reason we have things like `K"doc"` as a psuedo-keyword even though it's not exactly a keyword, `@doc` does still have special whitespace rules (unfortunately?! heh)

So my question is - did you consider just `bump_glue()` to glue the `K"."` tokens together (into `K"..."`, for example) rather than a `K"dots"` nonterminal?  To my mind that is somehow more appealing and feels similar in spirit to all the `remap_kind` we do, where we second guess the tokenizer quite a lot.

So - what makes `K"dots"` more appealing? Is it important that we try to delete `bump_glue()`?

c42f:
> but if we don't have to have it, I just fail to see a strong motivation to require it

Haha I actually find this the most compelling argument for removing it.

I do "like" token gluing in the sense that it allows `..` to be a single token, and that no other operators are more than one token, currently (I think?)

So removing gluing breaks this invariant of the output which we currently have.  Maybe that's not something to worry about though.

There might be other cases where it would be useful to create more rather than less tokens? For example, operator suffixes; `+₁` for example:  the tokenizer currently treats suffixes in one place independent from the operator start character, but then mashes them together with the operator start character into a single token. It's possible these would make sense as two separate tokens.

> to be able to test structural correctness of the parser

Seems reasonable - what kind of tests are you imagining?

> I also don't like using the same syntax head with two different meaning depending on whether it's a terminal or non-terminal

Yes I agree this feels weird.

An alternative I considered in the past is to have different kinds, eg for functions - `K"function"` and `K"function_token"` or something, but this breaks the property that the head names are "self-representing" in the way they are named in the source code and I didn't want to loose that.

A reasonable middle ground could be to have a new string macro so we'd have `K"function"` for the internal node and `KT"function"` for the token.

> Well, I don't really like the remapping either, but that's for a future PR.

Julia has many cases where a token might be a keyword or identifier depending on parser context. I don't love this, but I don't know what to do about it.

Keno:
> I do "like" token gluing in the sense that it allows `..` to be a single token, and that no other operators are more than one token, currently (I think?)

If it makes you feel better, you can think of `..` as dotted `.` ;). But yes, I agree the `..` thing is a little ugly, but it also seems like too much of a corner case to add a whole token gluing feature for.

> There might be other cases where it would be useful to create more rather than less tokens? For example, operator suffixes; +₁ for example: the tokenizer currently treats suffixes in one place independent from the operator start character, but then mashes them together with the operator start character into a single token. It's possible these would make sense as two separate tokens.

I actually want to go the other way and remove the operator-specific heads for these - WIP PR coming shortly.

> Seems reasonable - what kind of tests are you imagining?

Run the lexer, then run the parser, make sure the token ranges match. Validate that the total number of children matches the total number of nodes, validate the node summation invariants. This should catch the most obvious ways in which you could mess up the parser state - bumping incorrect things into the output or `emit`'ing something with a mark that isn't actually valid.

> Julia has many cases where a token might be a keyword or identifier depending on parser context. I don't love this, but I don't know what to do about it.

Keywords used as not-identifiers should always have the trivia flag set, no?

Keno:
I guess for this PR the other option is that we can always emit `...` as `K"."`, `K".."`, treat these as trivia and use the numeric flags for the ascension depth in importpath - that would let us retain `K".."` as a terminal while also getting rid of the splitting.

c42f:
> If it makes you feel better, you can think of .. as dotted . ;).

Haha!  No, I'm afraid it doesn't :laughing:

I feel think it's strictly worse to have `K"."` and `K".."` split in an ad hoc way which "just happens" to work.  If we're doing individual `K"."` it's nice and systematic to always just emit a sequence of them and have the parser group them as necessary.

> seems like too much of a corner case to add a whole token gluing feature for.

I agree. There's only one other case I think: negative integer literals.  Idk if those are essential either, maybe we could emit them as an interior node as well.

> I actually want to go the other way and remove the operator-specific heads for these - WIP PR coming shortly.

You mean this issue https://github.com/JuliaLang/JuliaSyntax.jl/issues/334 ?

I do think we should eliminate all the kinds which are inessential to parsing. The only things we'll have left are precedence classes of operators, plus a few specific operator kinds which are treated specially during parsing, such as `K"+"` for `parse_with_chains()`.

So I'm not suggesting adding more kinds for operators. The thought bubble here was that, perhaps, operators such as `^₁₁` might make sense to be tokenized as `"^"::PowerOperator "₁₁"::OpSuffix` or some such.  I don't know if this makes sense, though.

c42f:
Not to get too distracted about operators - I added a couple more thoughts about reducing the number of kinds into https://github.com/JuliaLang/JuliaSyntax.jl/issues/334#issuecomment-3060362304

c42f:
> >  Julia has many cases where a token might be a keyword or identifier depending on parser context. I don't love this, but I don't know what to do about it.
>
> Keywords used as not-identifiers should always have the trivia flag set, no?

Yes.

It's the opposite which is the problem.

Consider `outer = 100`

The tokenizer needs to emit `K"outer"`, or equivalent, so that this context-dependent keyword can be detected in cases when it's a keyword.

But when it's *not* used as a keyword, we remap it to `K"Identifier"`. The parser needs to make that decision and I don't think we can or should try to avoid that (the way to avoid this cleanly imo, it to just not allow such context-dependent keywords, but I guess that ship sailed a long time ago).

I don't know how to avoid this; we can shift the problem around, or use different terminology for it, but it seems unavoidable.  Maybe, maybe, we could insert extra tokenizer state or checks to try to get this correct for `outer`, but the parser already has all the context it needs, and complex tokenizer state hacks are very ugly and play badly with parser recovery.  (The stuff around tokenizing string interpolations is extremely evil IMO)

Keno:
> The tokenizer needs to emit `K"outer"`, or equivalent, so that this context-dependent keyword can be detected in cases when it's a keyword.

What I'm saying is that whenever a keyword is encountered as non-trivia, it should be treated as an identifier. We already have nodes other than K"Identifier" that are treated as identifiers (K"macro_name", K"var", this would add K".."), so whatever higher level API looks at these already needs to be aware of the possibility that there could be non-K"Identifier" identifiers, so why do we need to remap?

c42f:
> What I'm saying is that whenever a keyword is encountered as non-trivia, it should be treated as an identifier.

Ok, I understand.

But this further complicates the public API of the output data structure in order to satisfy an internal invariant - an invariant that I'm not yet convinced is particularly helpful (maybe you can convince me).

My point of view is that, during parsing, the parser has figured out that `K"outer"` is in Identifier-position and should only ever be treated as a normal `K"Identifier"` by downstream tooling. So that decision should be clearly and simply recorded. Otherwise we force the consumers of our API to figure this out again and cater to more special cases.

Keno:
> `K"Identifier"` by downstream tooling

What I'm saying is that any downstream tooling that explicitly looks at K"Identifier" is wrong because there are non K"Identifier" identifiers already (that we cannot remap). Now, conceivably we could restore the property that all identifiers are K"Identifier" and distinguish the various forms with flags, but that's not where we are right now.

c42f:
> What I'm saying is that any downstream tooling that explicitly looks at K"Identifier" is wrong because there are non K"Identifier" identifiers already

Somewhat agree.  Typically I write "big switch" style code to process this stuff

```
if k == K"Identifier"
 ....
elseif k == K"var"
...
...
end
```

In which case it's not wrong to refer to `K"Identifier"` explicitly.

But I appreciate the naming can be a trap for new consumers of the API: they could easily assume `K"Identifier"` is the only kind of thing which can be an identifier. It would be nice to make this trap harder to fall into.

Keno:
> But I appreciate the naming can be a trap for new consumers of the API: they could easily assume `K"Identifier"` is the only kind of thing which can be an identifier. It would be nice to make this trap harder to fall into.

My proposal would be to make `KSet"Identifier"` work and cover all identifier-like things and rename `K"Identifier"` to something like `K"PlainIdentifier"` or `K"RawIdentifier"` and tell people to use the former.

Keno:
Friendly ping @c42f - If you don't have an opinion, I think I'd like to merge as is.

c42f (commented):
Thanks for doing this, sorry I haven't looked back at it in a bit.

I really like most of the changes here, it's especially pleasing that 
* The tokenizer emits `K"."` only
*  `K"..."` is now only for splats
* We can delete `bump_split()` which is obviously terrible :laughing: 

I must admit I'm finding it hard to like `K"dots"`. I still feel like it exposes an implementation detail of the parser/tokenizer (ie, "It's hard to tokenize dots in Julia because they're used for lots of things") to the end user of syntax trees and I'm not really convinced about the benefit yet.

The theoretical benefit is that we might delete `bump_glue()` at some point and gain an invariant that we don't rewrite the width of tokens. But I don't know how this invariant is actually going to help us in practice?

Currently I'd rather prefer we just `bump_glue()` the dots together, and emit `K"Identifier"` rather than `K"dots"`. If you're really convinced `K"dots"` is the way to go, I feel it would be nice to rename it to `K"DotsIdentifier"` because that's more explicit.

Keno:
Alright, I will switch this to bump_glue and then open a revert of that as a PR. We can then decide whether to merge that later based on whether we think the invariant is important or not.

Keno:
I've addressed the other two review comments.

c42f (approved):
I don't love `orig_kind` (or `remap_kind`), but I see it as it's a symptom of Julia being impossible to tokenize neatly without involving the parser - dealing with this in some way or other feels like essential complexity. I do prefer to spare the users of `JuliaSyntax` from worrying about that, even at the cost of some ugly implementation things.

So I still prefer the `bump_glue` approach. Yes, we would have an internal `K".."` for this one specific edge case and I agree that's pretty annoying. But at least the user doesn't need to know, in contrast to having an extra kind `K"DotsIdentifier"` which they do need to know about. Surely that's an improvement?

Keno:
Merged upstream

PR #575, opened 2025-07-11
Remove separate syntax heads for each operator

### LINKED ISSUE #334: We don't need 800 kinds

Keno:
This replaces all the specialized operator heads by a single K"Operator" head that encodes the precedence level in its flags (except for operators that are also used for non-operator purposes). The operators are already K"Identifier" in the final parse tree. There is very little reason to spend all of the extra effort separating them into separate heads only to undo this later. Moreover, I think it's actively misleading, because it makes people think that they can query things about an operator by looking at the head, which doesn't work for suffixed operators.

Additionally, this removes the `op=` token, replacing it by two tokens, one K"Operator" with a special precendence level and one `=`. This then removes the last use of `bump_split` (since this PR is on top of #573).

As a free bonus this prepares us for having compound assignment syntax for suffixed operators, which was infeasible in the flips parser. That syntax change is not part of this PR but would be trivial (this PR makes it an explicit error).

Fixes #334

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -101,7 +101,6 @@ JuliaSyntax.is_infix_op_call
 JuliaSyntax.is_prefix_op_call
 JuliaSyntax.is_postfix_op_call
 JuliaSyntax.is_dotted
-JuliaSyntax.is_suffixed
 JuliaSyntax.is_decorated
 JuliaSyntax.numeric_flags
 ```
--- a/src/JuliaSyntax.jl
+++ b/src/JuliaSyntax.jl
@@ -41,7 +41,11 @@ export SourceFile
 @_public source_line_range
 
 # Expression predicates, kinds and flags
-export @K_str, kind
+export @K_str, kind, PrecedenceLevel, PREC_NONE, PREC_ASSIGNMENT,
+    PREC_PAIRARROW, PREC_CONDITIONAL, PREC_ARROW, PREC_LAZYOR, PREC_LAZYAND,
+    PREC_COMPARISON, PREC_PIPE_LT, PREC_PIPE_GT, PREC_COLON, PREC_PLUS,
+    PREC_BITSHIFT, PREC_TIMES, PREC_RATIONAL, PREC_POWER, PREC_DECL,
+    PREC_WHERE, PREC_DOT, PREC_QUOTE, PREC_UNICODE_OPS, PREC_COMPOUND_ASSIGN, generic_operators_by_level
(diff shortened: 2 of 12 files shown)


c42f (commented):
Ok, so I like this a lot in overview and I think the idea is right.

But there's a fair bit to clean up in the implementation and I'm going to be honest, this took an absolute ton of time to review.

One pervasive issue I find confusing in the parser.jl code changes is that `isassign` output of `peek_dotted_op_token()` is often ignored, but not always. Which cases is this actually ok for? One practical difference between ignoring `isassign` vs not is the difference between the following errors:

`isassign` checked for:
```julia
julia> parsestmt(SyntaxNode, "x |>= y")
ERROR: ParseError:
# Error @ line 1:3
x |>= y
# └┘ ── Compound assignment is not allowed for this operator
```

vs `isassign` not checked for:
```julia
julia> parsestmt(SyntaxNode, "x ..= y")
ERROR: ParseError:
# Error @ line 1:5
x ..= y
#   ╙ ── unexpected `=`
```

Was Claude or other AI tool used for the code changes?

I feel we may need a bunch more tests to check that `isassign` is used correctly.

Keno:
Thanks for the extensive review - I'll wait until the other PR is merged to rebase and fix those up.

> Was Claude or other AI tool used for the code changes?

I used claude off and on for the big changeset that this was extracted out of (along with several of the earlier PRs). However, I think the things you flagged as most objectionable are not Claude's fault, but rather an artifact of multiple versions of iterations and rebases as the earlier patches in this sequence were cleaned up and put in individually.

c42f:
Heya, we just moved JuliaSyntax into Base in https://github.com/JuliaLang/julia/pull/59870 - if you want to re-propose this PR over there, I've created a branch https://github.com/JuliaLang/JuliaSyntax.jl/tree/pr-for-Base/575 to make this easier.

* Clone JuliaLang/julia
* Check out pr-for-Base/575
* Rebase on top of master

For example:

```bash
# cd julia_dir

git fetch JuliaSyntax
git checkout pr-for-Base/575
git rebase origin/master
```

Keno:
Moved upstream.

PR #580, opened 2025-07-25
Fix multiline function signature parsing

KristofferC:
👨 @fredrikekre had an issue where the following parse was weird

```
julia> using JuliaSyntax

shell> cat bug.jl
function (
        ::A
        )()
end

julia> node = JuliaSyntax.parseall(JuliaSyntax.GreenNode, read("bug.jl", String))
     1:39     │[toplevel]
     1:38     │  [function]
     1:8      │    function
     9:9      │    Whitespace
    10:32     │    [tuple]
    10:10     │      (
    11:19     │      NewlineWs
    20:22     │      [::]
    20:21     │        ::
    22:22     │        Identifier       ✔
    23:31     │      NewlineWs
    32:32     │      )
    33:35     │    [block]
    33:34     │      [tuple]
    33:33     │        (
    34:34     │        )
    35:35     │      NewlineWs
    36:38     │    end
    39:39     │  NewlineWs
```

I let Claude lose on it and occording to it the issue was that with newlines the `peek(ps, 2)` check didn't function properly. 

This is a bit AI slop so it might not make sense to merge but it might point to where the issue is at least.

----------------

🤖 

Multiline function signatures with type annotations were incorrectly
parsed as tuples instead of calls when newlines appeared between
parentheses. For example:

```julia
function (
    ::A
)()
end
```

was parsed as `(function (tuple ...) (block))` instead of the correct
`(function (call (parens ...)) (block))`, inconsistent with the
single-line version `function (::A)() end`.

The issue was in parse_function_signature where `peek(ps, 2)` was used
to detect if a call pattern follows the closing parenthesis, but this
didn't skip newlines. 

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>


src/julia/parser.jl
```diff
@@ -2197,7 +2197,15 @@ function parse_function_signature(ps::ParseState, is_function::Bool)
             is_empty_tuple = peek(ps, skip_newlines=true) == K")"
             opts = parse_brackets(ps, K")") do had_commas, had_splat, num_semis, num_subexprs
                 _parsed_call = was_eventually_call(ps)
-                _needs_parse_call = peek(ps, 2) ∈ KSet"( ."
+                # Check if we should skip newlines - only for specific cases
+                # where we have a single type annotation like (::T)
+                _skip_newlines = !had_commas && num_subexprs == 1 &&
```
KristofferC, line 2202: I'll be honest in that I don't really know what is going on here...

Keno: I think the correct solution is to move the `_needs_parse_call` outside the do block - it seems weird that it would affect `needs_parameters`, because if there are parameters, the peek will never actually see the parenthesis.

fogti:
It might be the case that this breaks handling of newlines after the argument list.

PR #581, opened 2025-08-04
Abbreviated non-MIME `show()` for `GreenNode`

c42f:
Non-MIME `show()` for `GreenNode` was using the default implementation which is extremely verbose. Here we print it as an S-expression instead. I've made a choice to show absolute position within the parent node rather than node span so that it directly relates to indices in the string it was derived from.


src/porcelain/green_node.jl
```diff
@@ -118,6 +118,27 @@ function Base.show(io::IO, ::MIME"text/plain", node::GreenNode, str::AbstractStr
     _show_green_node(io, node, "", 1, str, show_trivia)
 end
 
+function _show_green_node_sexpr(io, node::GreenNode, position)
+    if is_leaf(node)
+        print(io, position, "-", position+node.span-1, "::", untokenize(head(node); unique=false))
+    else
+        print(io, "(", untokenize(head(node); unique=false))
+        first = true
```
aviatesk, line 126: This variable doesn't seem to be used at all?

c42f: Thanks, fixed.   Also fixed it in the `SyntaxNode` code this was derived from.

PR #586, opened 2025-08-25
`_register_kinds!`: prevent unintentional closure capture, boxing

nsajko:
The variable `i` was unintentionally shared between the generator closure in `_register_kinds!` and another part of the body of `_register_kinds!`. Thus `i` was boxed, causing trouble for inference.

Fix this by moving the part of `_register_kinds!` with `i` to a new function.

Fixing this should make the sysimage more resistant to invalidation, once the change propagates to Julia itself.

### DIFF
--- a/src/julia/kinds.jl
+++ b/src/julia/kinds.jl
@@ -102,6 +102,10 @@ function _register_kinds!(kind_modules, int_to_kindstr, kind_str_to_int, mod, mo
             error("Kind module ID $module_id already claimed by module $m")
         end
     end
+    _register_kinds_names!(int_to_kindstr, kind_str_to_int, module_id, names)
+end
+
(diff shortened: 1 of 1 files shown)


src/julia/kinds.jl
```diff
@@ -102,6 +102,10 @@ function _register_kinds!(kind_modules, int_to_kindstr, kind_str_to_int, mod, mo
             error("Kind module ID $module_id already claimed by module $m")
         end
     end
+    _register_kinds_names!(int_to_kindstr, kind_str_to_int, module_id, names)
+end
+
+function _register_kinds_names!(int_to_kindstr, kind_str_to_int, module_id, names)
```
aviatesk, line 108: Add a comment that explains why this separation is necessary, or use `let` or `local`.

PR #588, opened 2025-08-25
backport to v1: `_register_kinds!`: improve type stability when creating `Vector`

nsajko:
Make the temporary `existing_kinds::Vector` infer concretely.

NB: a more thorough solution might do away with the construction of `existing_kinds` altogether, however this seems like an OK quick fix.

(cherry picked from commit 27741256a1488c5016a3255afa76af56548cbe5d)

xref:

* the PR being backported: #587

* backport release PR #577

PR #589, opened 2025-08-25
backport to v1: `_register_kinds!`: prevent unintentional closure capture, boxing

nsajko:
The variable `i` was unintentionally shared between the generator closure in `_register_kinds!` and another part of the body of `_register_kinds!`. Thus `i` was boxed, causing trouble for inference.

Fix this by moving the part of `_register_kinds!` with `i` to a new function.

Fixing this should make the sysimage more resistant to invalidation, once the change propagates to Julia itself.

xref:

* the PR being backported: #586

* backport release PR #577


nsajko:
This PR seems to eliminate the exact same invalidations as PR #588.

nsajko:
bump

PR #607, opened 2025-12-01
Include missing sub- and superscripts in operator suffix list

moble:
As discussed in JuliaLang/julia#60285, the list of operator suffixes is missing several important sub- and superscripts — specifically, superscript alpha ᵅ, epsilon ᵋ, and (Latin) phi ᶲ, and subscript less than ˱, greater than ˲, and schwa ₔ.  This PR just adds those in the appropriate places.


moble:
@topolarity I'm confused about the purpose of this repo; it looks like there's still activity here?  So should I close this PR in favor of the one on julia itself?

Issue #600, opened 2025-10-16
`parse_brackets` should probably specialize on the function


KristofferC:
Saw the following invalidation:

<img width="1020" height="242" alt="Image" src="https://github.com/user-attachments/assets/aff6f5f8-cad4-4757-909b-b32c24738b30" />

JamesWrigley:
This is also causing invalidations in DataStructures/IJulia: https://github.com/JuliaLang/IJulia.jl/issues/1208#issuecomment-3482704095

JamesWrigley:
Also observed this to cause a ton of invalidations when loading CUDA.jl since that brings in CEnum.jl:
```julia
 inserting |(a::T, b::CEnum.Cenum{S}) where {T<:Integer, S<:Integer} @ CEnum ~/.julia/packages/CEnum/B6ahP/src/operators.jl:13 invalidated:
   mt_backedges: 1: signature Tuple{typeof(|), Int64, Any} triggered MethodInstance for Base.JuliaSyntax._register_kinds!(::Dict{Int64, Union{Module, Symbol}}, ::Dict{UInt16, String}, ::Dict{String, UInt16}, ::Module, ::Int64, ::Vector{String}) (0 children)
                 2: signature Tuple{typeof(|), Int64, Any} triggered MethodInstance for Base.JuliaSyntax._register_kinds!(::Dict{Int64, Union{Module, Symbol}}, ::Dict{UInt16, String}, ::Dict{String, UInt16}, ::Module, ::Int64, ::Vector{String}) (1 children)
                 3: signature Tuple{typeof(|), Bool, Any} triggered MethodInstance for <=(::Any, ::Char) (1 children)
                 4: signature Tuple{typeof(|), Bool, Any} triggered MethodInstance for <=(::Char, ::Any) (17 children)
                 5: signature Tuple{typeof(|), UInt16, Any} triggered MethodInstance for Base.JuliaSyntax.parse_function_signature(::Base.JuliaSyntax.ParseState, ::Bool) (1157 children)
```

(relevant issue: https://github.com/JuliaGPU/CUDA.jl/issues/2982)

JamesWrigley:
Bah humbug, #613 does not fix the issue on 1.12 ☹️ 
```julia
[...]

PR #615, opened 2025-12-21
Use structs instead of NamedTuple's in `parse_brackets()`

JamesWrigley:
@KristofferC, @mlechu, this is a more invasive but reliable (I think) solution to #600. Can confirm that it fixes the invalidations from `parse_brackets()` on 1.12 for InitialValues.jl (though there's a plethora of others). If y'all think it's a good idea I can make a PR to the main repo.

### DIFF
--- a/src/parser.jl
+++ b/src/parser.jl
@@ -1188,6 +1188,14 @@ function parse_juxtapose(ps::ParseState)
     end
 end
 
+@kwdef mutable struct ParseUnaryBracketOpts
+    delim_flags::RawFlags = 0
+    needs_parameters::Bool
+    is_paren_call::Bool
+    is_block::Bool
+end
+
+
 # Parse numeric literal prefixes, calls to unary operators and prefix
 # calls involving arbitrary operators with bracketed arglists (as opposed to
 # infix notation)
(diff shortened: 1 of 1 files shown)


src/parser.jl
```diff
@@ -1188,6 +1188,14 @@ function parse_juxtapose(ps::ParseState)
     end
 end
 
+@kwdef mutable struct ParseUnaryBracketOpts
```
c42f, line 1191: Couple of notes:
1. Why are these options structs mutable?
2. Unfortunately `@kwdef` doesn't exist on older julia versions

src/parser.jl
```diff
@@ -1282,9 +1290,9 @@ function parse_unary(ps::ParseState)
                             (initial_semi && num_subexprs > 0)    ||
                             (initial_semi && num_semis == 1)      ||
                             (num_semis == 0 && num_subexprs == 0)
-            return (needs_parameters=is_paren_call,
-                    is_paren_call=is_paren_call,
-                    is_block=!is_paren_call && num_semis > 0)
+            return ParseUnaryBracketOpts(; needs_parameters=is_paren_call,
+                                         is_paren_call=is_paren_call,
+                                         is_block=!is_paren_call && num_semis > 0)
         end
```
c42f, line 1296: Does this do the trick?

```suggestion
            return (needs_parameters=is_paren_call,
                    is_paren_call=is_paren_call,
                    is_block=!is_paren_call && num_semis > 0)
        end
        opts::NamedTuple{(:needs_parameters, :is_paren_call, :is_block), Tuple{Bool, Bool, Bool}}
```

c42f (commented):
I think this is fine if we need it.  Though I'd prefer a less invasive alternative if we want to keep the code idiomatic for future Julia versions (assuming 1.13 doesn't suffer from this?)

The simplest least invasive way could be to type assert the type of opts as soon as it's returned?  Then the performance hacking is localized in one place?

